PR #134, opened 2015-01-23
New Payment model and repayments

vivekraghuram:
Repayments is used to refer to payments made by an org on a loan.


noahsark769:
Aww y u close dis? I was just gunna look at it haha

vivekraghuram:
haha except now i need to change some things about how I did it.

Issue #111, opened 2015-01-19
New styles for login page


noahsark769:
The forms are way to big now, we have to figure out a revolv-y way to display them. Definitely could have smaller forms on desktop styled with thin Helvetica Neue like the homepage is, then the forms could be bigger on mobile

noahsark769:
#143

PR #143, opened 2015-02-10
New styles for login page

noahsark769:
For #111. I'm excited about this, they're so much better than the old ones and they animate between login/signup! I'd like at least one review before merging this so can someone look over it? Maybe @sameerav or @aykamko 

screen shot 2015-02-10 at 1 41 06 am
screen shot 2015-02-10 at 1 41 15 am
screen shot 2015-02-10 at 1 41 31 am
screen shot 2015-02-10 at 1 41 44 am


sameerav:
I like these a lot! I don't know if it does this right now but it might look cool to have the login button fill in when you hover above it

Also, for the case where the user didn't enter a username or password, it might be better to just highlight the box - the full red fill is a little overwhelming

noahsark769:
Yup, it fills in the box on hover! Also yeah, let me are if I can get it to
look good with just a red border

andylouisqin:
These look slick @noahsark769! Stamp of approval.

noahsark769:
Thanks @andylouisqin! @sameerav thanks for the suggestion on highlighting with border instead, I think it looks much nicer. What do you think?

screen shot 2015-02-10 at 10 40 17 am

PR #144, opened 2015-02-12
Repayment

vivekraghuram:
Repayment!


revolv/payments/migrations/0003_auto_20150103_0323.py
```diff
@@ -0,0 +1,23 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def populate_data(apps, schema_editor):
+    PaymentInstrumentType = apps.get_model("payments", "PaymentInstrumentType")
+
+    PaymentInstrumentType.objects.create(name="reinvestment")
+    PaymentInstrumentType.objects.create(name="check")
```
noahsark769: So the "check" type is for someone who made a manual donation to a project via check right? We should document somewhere (PaymentInstrumentType docstring?) what kinds of payment instrument types there are and what their functions are

revolv/payments/models.py
```diff
@@ -11,7 +12,19 @@ def get_paypal(self, queryset=None):
         """Return the PaymentInstrumentTypeManager for paypal payments."""
         if queryset is None:
             queryset = super(PaymentInstrumentTypeManager, self).get_queryset()
-        return queryset.get(name="paypal")
+        return queryset.get(name=INSTRUMENT_PAYPAL)
+
+    def get_reinvestment(self, queryset=None):
+        """Return the PaymentInstrumentTypeManager for paypal payments."""
```
noahsark769: You mean reinvestment payments, right?

revolv/payments/models.py
```diff
@@ -23,63 +36,92 @@ class PaymentInstrumentType(models.Model):
     objects = PaymentInstrumentTypeManager()
 
 
-class PaymentTransactionManager(models.Manager):
+class PaymentManager(models.Manager):
+    """
+        Simple manager for the Payment model.
+    """
+
+    def payments(self, user=None, project=None, queryset=None):
```
noahsark769: Why include project in the kwargs if we don't use it?

revolv/payments/models.py
```diff
@@ -23,63 +36,92 @@ class PaymentInstrumentType(models.Model):
...
+        """
+        :return: Payments associated with this user
+        """
+        if queryset is None:
+            queryset = super(PaymentManager, self).get_queryset()
+        payments = queryset.filter(user=user).order_by('created_at')
+        return payments
+
+    def donations(self, user=None, project=None, queryset=None):
```
noahsark769: There's a lot of repeated code in these methods, couldn't we extract some of it?

revolv/payments/services.py
```diff
@@ -13,12 +15,14 @@ class PaymentServiceException(Exception):
 class PaymentService(object):
 
     @classmethod
-    def create_payment(cls, user, amount, payment_instrument):
+    def create_payment(cls, user, entrant, amount, project, payment_instrument):
         """
         Create a payment based on a configured payment_instrument.
 
         :user: a User making the payment
+        :entrant: a User entering the payment
```
noahsark769: Can we add something about that if the user is the same as the entrant, it represents an organic donation?

revolv/payments/tests/models/test_payment.py
```diff
@@ -0,0 +1,112 @@
...
+from django_facebook.utils import get_user_model
+from revolv.base.models import RevolvUserProfile
+from revolv.base.signals import create_profile_of_user
+from revolv.payments.models import Payment, PaymentInstrumentType
+from revolv.project.models import Project
+
+
+class PaymentTest(TestCase):
+
+    def setUp(self):
```
noahsark769: We could do like `self.reinvestment = PaymentInstrumentType.objects.get_reinvestment()` to save a little verbosity here: we're doing `PaymentInstrumentType.objects.get_<something>()` a lot in these tests haha

revolv/payments/tests/models/test_payment_instrument_type.py
```diff
@@ -1,12 +1,16 @@
...
 class PaymentInstrumentTypeTest(TestCase):
 
-    fixtures = ['supported_instruments.json']
-
     def test_get_payment_instrument_types(self):
 
-        self.assertEqual(1, len(PaymentInstrumentType.objects.all()))
+        self.assertEqual(4, len(PaymentInstrumentType.objects.all()))
         self.assertEqual(INSTRUMENT_PAYPAL, PaymentInstrumentType.objects.first().name)
+        self.assertEqual(INSTRUMENT_REINVESTMENT, PaymentInstrumentType.objects.get(id=2).name)
```
noahsark769: Are you sure it's a good idea to test on the the specific ids? Technically there's no guarantee that they'll have these ids in any database right, which could make this test fail in random places

revolv/payments/tests/services/test_payment_service.py
```diff
@@ -30,17 +33,35 @@ def test_create_payment(self, mock_amount, mock_instrument):
...
+        try:
+            PaymentService.create_payment(
+                user,
+                None,
+                10.00,
+                project,
+                payment_instrument,
+            )
+        except PaymentServiceException:
+            failed = True
```
noahsark769: Instead of having the boolean you can just use `self.fail()`

revolv/project/models.py
```diff
@@ -286,9 +282,9 @@ def amount_donated(self):
         :return: the current total amount that has been donated to this project,
             as a float
         """
-        result = self.donation_set.aggregate(
-            models.Sum('payment_transaction__amount')
-        )["payment_transaction__amount__sum"]
+        result = Payment.objects.donations(project=self).aggregate(
```
noahsark769: +1, this is a much more robust way to do repayments!

revolv/templates/base/partials/nav.html
```diff
@@ -26,7 +26,13 @@
                <li class="user-controls has-dropdown">
                    <a href="#">{{user.username}}</a>
                    <ul class="dropdown">
+                       {% if is_administrator %}
```
noahsark769: The thing about the dashboard going to the highest level was already fixed in #140, so we can just make this like go to "/dashboard/" and I believe it should be fine. @aykamko can confirm

noahsark769:
This looks really good!! I'm so excited it's finally getting there haha. Looks like master needs to be merged in but other than that I say ship it after fixing a few small things I mentioned

revolv/payments/migrations/0003_auto_20150103_0323.py
vivekraghuram: Agreed. Start adding to the Google doc or readme?

revolv/payments/models.py
vivekraghuram: @noahsark769 it's there for consistency with the other manager functions.

revolv/payments/tests/models/test_payment_instrument_type.py
vivekraghuram: @noahsark769 the IDs work mainly because those values are in there because of a data migration so we know what ID they should be. Nothing else should write to that table. If this test breaks than something must be wrong.

revolv/payments/migrations/0003_auto_20150103_0323.py
noahsark769: I was thinking either a docstring or the readme, cause it's directly related to the code. And if readme, I think we should have a reference somewhere in the docstring for PaynentTransactionType anyway

revolv/payments/models.py
noahsark769: Hm fair enough I guess

revolv/payments/tests/models/test_payment_instrument_type.py
noahsark769: Hmmm true...but technically the create method doesn't guarantee the ids to be 1,2,3 etc...though I guess you're right that if this test breaks something is wrong so I think this is okay

revolv/payments/models.py
```diff
@@ -23,63 +36,92 @@ class PaymentInstrumentType(models.Model):
...
+        """
+        :return: Returns all payments that are not repayments or reinvestments
+                 associated with this user.
+        """
+        if queryset is None:
+            queryset = super(PaymentManager, self).get_queryset()
```
SamLau95: Do you think it's cleaner to write:

    queryset = queryset or super(PaymentManager, self).get_queryset()

revolv/payments/models.py
SamLau95: Could you link me to the other manager functions you talk about?

If you were to DRY it out, would it be possible to have a private method called something like  `makeQuery` that each of these methods could just delegate to (call with right arguments)?

revolv/payments/services.py
```diff
@@ -13,12 +15,14 @@ class PaymentServiceException(Exception):
 class PaymentService(object):
 
     @classmethod
-    def create_payment(cls, user, amount, payment_instrument):
+    def create_payment(cls, user, entrant, amount, project, payment_instrument):
```
SamLau95: This method has a lot of arguments; do you think it's worth making a small class / named tuple to pass this stuff around?

revolv/payments/tests/models/test_payment_instrument_type.py
SamLau95: If you do leave the IDs in, it might be worth considering making small functions/constants to abstract out the ID so the test can read something like

    self.assertEqual(INSTRUMENT_REINVESTMENT, reinvestment_payment().name)

Which is a little clearer to grok on a quick pass through the test.

revolv/project/models.py
```diff
@@ -311,6 +306,16 @@ def amount_left(self):
...
 
     @property
+    def amount_repaid(self):
+        """
+        :return: the current amount of money repaid by the project to RE-volv.
+        """
+        amount_repaid = Payment.objects.repayments(project=self).aggregate(models.Sum('amount'))["amount__sum"]
+        if amount_repaid is None:
+            return 0.0
+        return amount_repaid
```
SamLau95: Can this be simplified to

    return amount_repaid or 0.0

revolv/payments/models.py
noahsark769: He just means all the methods of `PaymentManager`. Tbh I think @SamLau95 is right, this code could be dried up with a a delegation function

revolv/payments/tests/models/test_payment_instrument_type.py
noahsark769: ^ true

revolv/payments/models.py
noahsark769: I would actually argue that the two line if statement rather than the one line or is easier to understand from a code readability perspective

SamLau95: Yeah this is contestable - the or equals pattern is very common in Ruby so
that's why I mention it.
ᐧ

On Sat, Feb 21, 2015 at 1:32 PM, Noah Gilmore <notifications@github.com>
wrote:

> In revolv/payments/models.py
> <https://github.com/calblueprint/revolv/pull/144#discussion_r25126036>:
>
> > +    def payments(self, user=None, project=None, queryset=None):
> > +        """
> > +        :return: Payments associated with this user
> > +        """
> > +        if queryset is None:
> > +            queryset = super(PaymentManager, self).get_queryset()
> > +        payments = queryset.filter(user=user).order_by('created_at')
> > +        return payments
> > +
> > +    def donations(self, user=None, project=None, queryset=None):
> > +        """
> > +        :return: Returns all payments that are not repayments or reinvestments
> > +                 associated with this user.
> > +        """
> > +        if queryset is None:
> > +            queryset = super(PaymentManager, self).get_queryset()
>
> I would actually argue that the two line if statement rather than the one
> line or is easier to understand from a code readability perspective
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/calblueprint/revolv/pull/144/files#r25126036>.
>



-- 
Sam Lau
EECS | UC Berkeley | May 2017
(562) 565-4336 | samlau95@berkeley.edu

revolv/payments/services.py
noahsark769: I don't think so - if we changed this method to accept a namedtuple instead, then we would just have to instantiate a namedtuple with all the arguments somewhere else

SamLau95: True - I don't know if that's worth it in this case but in general it's a code smell when a method has a bunch of parameters
http://blog.codinghorror.com/code-smells/
http://stackoverflow.com/questions/174968/how-many-parameters-are-too-many

revolv/payments/tests/services/test_payment_service.py
vivekraghuram: Won't this cause the test to fail?

noahsark769: what I mean to say is that whenever you set `failed = True`, you should just call `self.fail()` instead

vivekraghuram: Yeah but I want to make sure that the exception is caught. Won't calling self.fail() result in the entire test failing?

revolv/payments/models.py
vivekraghuram: now we do!

revolv/payments/tests/models/test_payment_instrument_type.py
noahsark769: I think there was a `get_paypal()` method around here somewhere that I wrote for this reason: we can go off that with `get_reinvestment()` and `get_repayment()`

revolv/payments/tests/services/test_payment_service.py
noahsark769: Oh I see. Can't you just use `assertRaises` instead?

revolv/payments/tests/models/test_payment_instrument_type.py
noahsark769: Oh wait we already have those, nice

noahsark769:
Awww yeeaaaahhhhh time to push the big green button

Issue #146, opened 2015-02-14
Sam doesn't know Django


SamLau95:
But I'm reviewing your PRs this week. Can I drop by your meeting this Sunday to go over the basics? :)

noahsark769:
Absolutely!

PR #147, opened 2015-02-15
Vagrant environment

noahsark769:
I figured we should have one of these, if only so @SamLau95 can get set up easily with regards to #146. I know @andylouisqin was having installation issues also so hopefully this should help. If anyone wants to look over this, go for it! Probably will merge it at the meeting tomorrow.


noahsark769:
Merging this for @sameerav

PR #156, opened 2015-02-18
added django_extensions, which lets us generate object model diagrams

aykamko:
Requires graphviz and pygraphviz
    
    brew install graphviz
    pip install pygraphviz


noahsark769:
Do we need this? Is there at least a way to make it only generate a graph with the models from our apps instead of django_facebook, django_cms, etc

noahsark769:
(When I say do we need this, I mean wouldn't it really just be simpler to draw out the database diagram that we need by hand? That graph is so huge and it seems like overkill)

SamLau95:
It is pretty hard for me to know where to start looking for this visualization ; is there a way to ignore models?

noahsark769:
^ that's what im sayin

noahsark769:
Okay imma close this, we can revisit it if we really need to

PR #162, opened 2015-02-24
Admin list emails

erictshen:
Added means for admin to acquire list of emails, and updated admin dashboard to have more than one page.

Also updated git ignore very quickly for some static files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,7 @@ docs/_build/
 raw_templates/node_modules/
 raw_templates/bower_components/
 revolv/static/bower_components/
+revolv/static/revolv/static/
 node_modules/
 .revolv_bash_profile
 .vagrant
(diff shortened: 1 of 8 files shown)


erictshen:
What second layer of tabs looks like. Let me know any feedback. I think it could be better, but not sure how to design it better.

screen shot 2015-02-23 at 10 10 28 pm

erictshen:
The pull request is currently not completed, will update it soon!

noahsark769:
I think the nested tabs could definitely at least be a little bit smaller to differentiate them, let me know when it's done and I'll check it out!

noahsark769:
Wait what would be in revolv/static/revolv/static?

erictshen:
Deleting this pull request and making a new one b/c I messed up git stuff...

PR #164, opened 2015-03-01
Admin list emails merged

erictshen:
Added means for admin to acquire list of emails, and updated admin dashboard to have more than one page.

============================
Same as the older one, but deleted it b/c I had git issues.

### DIFF
--- a/revolv/administrator/views.py
+++ b/revolv/administrator/views.py
@@ -1,9 +1,10 @@
 from itertools import chain
 
 from django.views.generic import TemplateView
+
+from revolv.base.models import RevolvUserProfile
 from revolv.base.users import UserDataMixin
 from revolv.project.models import Project
-from revolv.base.models import RevolvUserProfile
 
 
 class AdministratorDashboardView(UserDataMixin, TemplateView):
--- a/revolv/administrator/views.py
+++ b/revolv/administrator/views.py
@@ -8,7 +8,7 @@ from revolv.project.models import Project
 
 
 class AdministratorDashboardView(UserDataMixin, TemplateView):
-    """Basic view for the Administrator dashboard.
+    """Basic view for the Administrator dashboard. Shows the list of projects.
     """
     template_name = 'administrator/dashboard.html'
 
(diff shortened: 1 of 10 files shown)


erictshen:
What second layer of tabs looks like. Let me know any feedback. I think it could be better, but not sure how to design it better.

screen shot 2015-02-23 at 10 10 28 pm

erictshen:
From old pull request : Noah : I think the nested tabs could definitely at least be a little bit smaller to differentiate them, let me know when it's done and I'll check it out!

erictshen:
From old pull request : Noah : Wait what would be in revolv/static/revolv/static?

erictshen:
Finalized look.

screen shot 2015-03-01 at 1 28 33 pm

erictshen:
Updated view

screen shot 2015-03-01 at 1 31 33 pm

revolv/base/models.py
```diff
@@ -20,6 +20,19 @@ def create_user_as_admin(self, *args, **kwargs):
...
 
+    def get_subscribed_to_newsletter(self, queryset=None):
+        """ Gets all the user emails that are currently subscribed to the newsletter
+
+        :queryset: The queryset in which to search for users
+        :return: A list of users
+        """
+        if queryset is None:
+            queryset = super(RevolvUserProfileManager, self).get_queryset()
+            subscribed_users = queryset.filter(
```
noahsark769: De-indent :)

revolv/base/tests/tests.py
```diff
@@ -132,6 +136,22 @@ def test_signup_endpoint(self):
         RevolvUserProfile.objects.get(user=test_user)
 
 
+class RevolvUserProfileManagerTestCase(TestCase):
+    """Tests for the RevolvUserProfileManager"""
```
noahsark769: Need a todo to make test objects instead of fixtures

revolv/templates/administrator/email.html
```diff
@@ -0,0 +1,42 @@
...
+    <div class='small primary administrator-button button'>Ambassadors</div>
+    <a href="{%url "administrator:email" %}">
+        <div class='small active administrator-button button'>Emails</div>
+    </a>
+</div>
+
+<div class='row'>
+    <div class="small-12 columns">
+        <h4>Newsletter Subscribers</h4>
+        <textarea rows="15">{% for email in subscribed_user_emails %}{{ email|safe }}&#13;&#10;{% endfor %}</textarea>
```
noahsark769: This is an XSS injection opportunity since the user specified their email, so we should not use `|safe` :)

noahsark769: Or maybe it wont be a problem since we validate emails, but always better to be overly secure rather than underly secure

revolv/administrator/views.py
```diff
@@ -19,4 +21,16 @@ def get_context_data(self, **kwargs):
                                              context['active_projects'], context['completed_projects']))
         if len(context['all_projects']) > 0:
             context['active_project'] = int(self.request.GET['active_project']) if 'active_project' in self.request.GET else context['all_projects'][0].id
+        context['subscribed_user_emails'] = RevolvUserProfile.objects.get_subscribed_to_newsletter()
+        return context
+
+
+class AdministratorEmailView(UserDataMixin, TemplateView):
+    """Basic view for the Administrator dashboard.
```
noahsark769: Needs to not be copypasta :)

noahsark769:
YAY

PR #166, opened 2015-03-02
Added post donation email hook

### LINKED ISSUE #123: Post donation email hook
See description in #51. For this issue, let's just add in the hook that emails the person and the actual email, no need for html emails.

erictshen:
Email styling coming later!

Fixes #123

### DIFF
--- a/revolv/project/tests/tests_general.py
+++ b/revolv/project/tests/tests_general.py
@@ -1,10 +1,11 @@
 import datetime
 import json
-import mock
 
+import mock
 from django.core.management import call_command
 from django.db.models.signals import post_save
 from django.test import TestCase
+
 from django_facebook.utils import get_user_model
 from revolv.base.models import RevolvUserProfile
 from revolv.base.signals import create_profile_of_user
--- a/revolv/project/tests/tests_general.py
+++ b/revolv/project/tests/tests_general.py
@@ -261,9 +261,9 @@ class DonationAjaxTestCase(CreateTestProjectMixin, TestUserMixin, TestCase):
         self.project.project_status = Project.ACTIVE
         self.project.save()
 
-    def test_valid_donation(self):
+    def perform_valid_donation(self):
         """
-        Test valid donation via AJAX to /project/<pk>/donation/submit.
+        Utility method for performing a valid donation. Returns the response.
         """
         valid_donation = {
             'csrfmiddlewaretoken': self.client.cookies['csrftoken'].value,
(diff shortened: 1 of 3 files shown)


SamLau95:
@generic ++

noahsark769:
Looks good! Would you be able to write a little more for the email, keeping in mind that this is what RE-volv will want their users to see after donating and that we want to use this email to foster their future engagement? I feel like the text is pretty bland for those purposes now.

Also, can we write a test to make sure the email gets sent? You can probably write a simple one by mocking out django's email functionality and checking that mail.send is called, and we could even put that in a test utils file for later use

revolv/project/tests/tests_general.py
```diff
@@ -284,6 +286,33 @@ def test_valid_donation(self):
         self.assertIsNone(content.get('error'))
         self.assertIsNotNone(self.project.donors.get(pk=self.test_user.pk))
 
+    @mock.patch('revolv.lib.mailer.EmailMultiAlternatives')
+    def test_post_donation_email(self, mock_mailer):
+        """
+        Tests whether a valid donation will send a revolv email
+        """
+        valid_donation = {
```
noahsark769: This is in two tests now, can we pull it out to a utility?

revolv/project/tests/tests_general.py
```diff
@@ -284,6 +286,33 @@ def test_valid_donation(self):
...
+        resp = self.client.post(
+            self.project.get_absolute_url() + self.DONATION,
+            data=valid_donation,
+            HTTP_X_REQUESTED_WITH='XMLHttpRequest'
+        )
+        self.assertEqual(resp.status_code, 200)
+        content = json.loads(resp.content)
+        self.assertIsNone(content.get('error'))
+        self.assertIsNotNone(self.project.donors.get(pk=self.test_user.pk))
+        self.assertTrue(mock_mailer.called)
```
noahsark769: We can check assertCalledWith their email address also, I feel like that would be good

noahsark769:
The new text looks great!

revolv/project/tests/tests_general.py
erictshen: assertCalledWith also cares about the other four arguments, so did something like this, because we only care about the to email address
http://stackoverflow.com/questions/20428750/pythons-assert-called-with-is-there-a-wildcard-character

noahsark769: Oh nice!

noahsark769:
Awesome work!!

Issue #171, opened 2015-03-03
Signin and Signup are not clickable on the project page when not logged in


noahsark769:

noahsark769:
#174

Issue #172, opened 2015-03-03
When you close the payment modal, the credit card icons show briefly after the modal dissapears


noahsark769:
Seems to happen on staging all the time, when you click the first close button

noahsark769:
#174

PR #174, opened 2015-03-03
fixed display:table thing for payment modal

aykamko:


revolv/templates/base/base.html
```diff
@@ -61,6 +61,19 @@
           reveal: {
               animation: 'fade',
               dismiss_modal_class: 'close-revolv-reveal-modal',
+              css: {
+                  open: {
+                      'opacity': 0,
+                      'visibility': 'visible',
+                      // NOTE: you should manually change display of your modal
+                      // in the scss
```
noahsark769: probably we should explain why, if we're gunna have the comment

noahsark769:
This fixes #171 and #172 only, right?

aykamko:
Yeah, just those two. I'll expand the comment

noahsark769:
kk cool

PR #177, opened 2015-03-06
BIG testing refactor

### LINKED ISSUE #114: Ambassador can see Django CMS toolbar
This happens on staging. I don't think it should?

### LINKED ISSUE #173: Find some code pattern to not have to load in fixtures in tests
Loading in fixtures in tests is not a very good practice. We have a `CreateTestProjectMixin`...why not have that for all the models?

noahsark769:
Fixes #114 and #173 

It started out as a fix for the ambassador seeing the Django CMS toolbar...

1. In order to do this, I had to ass `CMS_PERMISSION` in settings.py.
2. When `CMS_PERMISSION` is enabled, django-cms hooks up a bunch of signals to user creation
3. Because of this, loading the same users through fixtures multiple times caused the hooks to run multiple times while only keeping the three users
4. This created empty database relationships which caused `IntegrityErrors` in lots of the tests.
5. I figured out that in order to make this work, I would have to disconnect and reconnect three more django-cms signals in every test that loads user fixtures. So at that point, I decided to just try to resolve #173 as well and take out all fixtures from tests
6. I put in [Factory Boy](http://factoryboy.readthedocs.org/en/latest/introduction.html) and used some magic to create a factory pattern for models. Different kinds of model factories are available through `.factories`: for example, `RevolvUserProfile.factories.admin.create()` creates a RevolvUserProfile that is an admin with default everything. If you want to specify it as a field, you pass it in as a kwarg e.g. `Project.factories.base.create(amount=50.0, org_name="Commoonity Dairy")`.
7. I refactored a BUNCH of tests to use this instead of fixtures. I also took out `CreateTestProjectMixin` because Factory Boy provides the exact same functionality but better.

Also, it may just be me but the tests seem to run a tiny bit faster now!

This is a pretty significant change in terms of our testing patterns so I'm requesting a shipit from @SamLau95 and @vivekraghuram before I merge this. Review from @erictshen and @aykamko would also be helpful. Thanks guys!


SamLau95:
This looks super legit! @noah++

erictshen:
@noah++ Looks very solid!

vivekraghuram:
This is amazing! @noah++ looks good to me.

vivekraghuram:
And so many fewer imports! revolv++

noahsark769:
Yeee alright imma merge this then

noah:
On Sat, Mar 07, 2015 at 11:05:21AM -0800, Vivek Raghuram wrote:
> Date: Sat, 07 Mar 2015 11:05:21 -0800
> From: Vivek Raghuram <notifications@github.com>
> To: calblueprint/revolv <revolv@noreply.github.com>
> Cc: "Noah K. Tilton" <noah@tilton.co>
> Subject: Re: [revolv] BIG testing refactor (#177)
> List-ID: calblueprint/revolv <revolv.calblueprint.github.com>
> 
> This is amazing! @noah++ looks good to me. 

What?

PR #195, opened 2015-03-11
Disallow opening the donate modal when the user is not logged in

### LINKED ISSUE #170: Donate modal should not appear for users who are not logged in
Now it does and that causes errors because anonymous users have no revolvuserprofile

noahsark769:
Fixes #170, also:

1. Add an integration `WebTest` to check to make sure the donate button redirects to the login page
2. Changes project and home pages to not display a cover photo if none is available. Had to do this because we can't populate cover photos in the tests, and Django will 400 if we try to access a cover photo when it's None.
3. Added some stuff so that when you go to the login page being redirected from the donate button (aka when not logged in), you'll see a "please log in to donate" instead of "welcome back", which I figured is a nice touch.

@aykamko since you know about donations can you give this a quick look please :D


revolv/templates/project/project.html
```diff
@@ -24,7 +24,11 @@
         {% endif %}
         <ul class="button-group nav-tab-button-group">
             {% if project.is_active %}
-                <li><a class="small button success radius" href="#donate" data-reveal-id="donate-modal">Donate</a></li>
+                {% if user.is_authenticated %}
+                    <li><a id="donate-button" class="small button success radius" href="#donate" data-reveal-id="donate-modal">Donate</a></li>
+                {% else %}
+                    <li><a id="donate-button" class="small button success radius" href="/signin/?next=/project/{{project.pk}}/&reason=donate#login">Donate</a></li>
+                {% endif %}
```
aykamko: I think this could be cleaner with a variable? Something like 

```
{% if user.is_authenticated %}
{% set donate_href = '#donate' %}
{% else %}
...
{% endif %}
<li><a id="donate-button" class="small button success radius" href="{{donate_href}}" data-reveal-id="donate-modal">Donate</a></li>
```

revolv/templates/base/sign_in.html
```diff
@@ -23,8 +23,14 @@
...
     <div class='component-container medium-offset-2 medium-4 small-12 columns'>
-        <h1>Welcome back!</h1>
-        <p>Log in to see the impact you've had on communities that use renewable solar energy.</p>
+        {% if reason and reason == "donate" %}
+            <h1>Log in to donate</h1>
+        {% else %}
+            <h1>Welcome back!</h1>
+        {% endif %}
+        {% if not reason or reason == "donate" %}
+            <p>Log in to see the impact you've had on communities that use renewable solar energy.</p>
```
aykamko: Can't this be moved into the above `{% else %}`?

noahsark769: yup

Issue #183, opened 2015-03-09
When you try to GET a drafted project, make the app 404 instead of 301


noahsark769:
It's confusing for someone to see that thy don't have access to a page if they try to request a project page that is drafted.

Issue #180, opened 2015-03-08
Add in "project tagline"


noahsark769:
According to the new project page design in #178, we need a short tagline for each project. We should add this in. And then if the project doesn't have a tagline, we should just not render the heading.

PR #197, opened 2015-03-11
A few small issue fixes

noahsark769:
1. Updated mailer api to be a little better: `send_revolv_email` now accepts kwargs to specify if it should fail silently or not and it returns the same return code as `EmailMultiAlternatives.send`.
2. Made the app 404 instead of redirect for pages of projects which are drafted (#183)
3. Added taglines to projects (#180)

Also I think something about my git is screwed up or pre-commit isn't working since I had to commit some stuff twice. More commits I guess ¯\_(ツ)_/¯

@erictshen can I get a quick review of this?


revolv/lib/mailer.py-e
```diff
@@ -0,0 +1,70 @@
+import yaml
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives
+from django.template import Context, Template
+from revolv.base.utils import get_all_administrator_emails
+
```
erictshen: I think this file got duplicated, make sure to not include it in the pull!

revolv/project/models.py
```diff
@@ -146,6 +145,12 @@ class Project(models.Model):
         max_length=255,
         help_text='How would you like to title this project?'
     )
+    tagline = models.CharField(
+        max_length=100,
+        null=True,
+        blank=False,
+        help_text='Select a short tag line that describes this project. (No more than 100 characters.)'
```
erictshen: I think you need blank=True in order for it to not be required.

https://docs.djangoproject.com/en/1.7/ref/models/fields/

erictshen:
LGTM, make sure to not include the extra mailer file and feel free to then ship it!

revolv/lib/mailer.py-e
noahsark769: Aw god dammit, why does it keep doing that

noahsark769:
Sweet! Thanks man

PR #200, opened 2015-03-15
Eric email csv

erictshen:
Allows the admin to download the emails in csv format. Added a "Download as CSV" button to administrator:email dashboard as well.
screen shot 2015-03-15 at 12 10 29 pm
screen shot 2015-03-15 at 12 10 39 pm


revolv/administrator/views.py
```diff
@@ -31,5 +33,23 @@ class AdministratorEmailView(UserDataMixin, TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super(AdministratorEmailView, self).get_context_data(**kwargs)
-        context['subscribed_user_emails'] = RevolvUserProfile.objects.get_subscribed_to_newsletter()
+
+        user_emails = [r.user.email for r in RevolvUserProfile.objects.get_subscribed_to_newsletter()]
```
noahsark769: Doesn't `get_subscribed_to_newsletter` return a `QuerySet`? If so, we could just to `.get_subscribed_to_newsletter().values("user__email")` instead of this list comprehension right?

revolv/administrator/views.py
```diff
@@ -31,5 +33,23 @@ class AdministratorEmailView(UserDataMixin, TemplateView):
...
+        context['subscribed_user_emails'] = user_emails
         return context
+
+
+def admin_email_csv_download(request):
```
noahsark769: This view could use a docstring :)

revolv/base/models.py
```diff
@@ -21,7 +22,9 @@ def create_user_as_admin(self, *args, **kwargs):
         return profile
 
     def get_subscribed_to_newsletter(self, queryset=None):
```
noahsark769: Oh looks like this does return a `QuerySet`. Would be helpful to have a `:return:` saying that it does also

noahsark769:
Looks great!

PR #201, opened 2015-03-15
refactor of payments models

### LINKED ISSUE #30: Repayment and Reinvestment Backend Refactor
Each project has an associated IRR (Internal Rate of Return). We need to use that IRR to calculate return on investment. Each month they will get some percentage of money back. Some of that money goes to revolv and the rest goes into a pool that can be reinvested. Suppose they get $100 back. $7 might go to revolv and then the remaining $93 could be reinvested. 

I was thinking there could be a new payment object, revolv_bank, which would keep track of how much re-volv itself has to reinvest. The revolv_bank could then make its own donation towards a project, just like any other payment, allowing money to be reinvested. 

Here's the tricky part: The reinvestment by revolv_bank needs to tie back to the donors some how so they can see how their money propagated through the system. So if Person A donates to project 1 and then revolv_bank donates the ROI on project 1 to project 2, Person A should see some of their impact on project 2. This gets more complicated by the fact that Person B could donate to project 2. If the returns from project 2 are reinvested in project 3, Person A should see impact on projects 1, 2, and 3 and Person B should see impacts on projects 2 and 3.

aykamko:
EDIT by Noah: For reference this fixes #30


aykamko:
Don't merge this! Just made a pull request so we can track progress via diff

noahsark769:
Aww yeah revolv++

aykamko:
yas revolv++

revolv/payments/models.py
```diff
@@ -1,64 +1,235 @@
...
+        that has this PaymentType.
     """
     name = models.CharField(max_length=80, unique=True)
 
-    objects = PaymentInstrumentTypeManager()
+    objects = PaymentTypeManager()
+
+
+class RepaymentManager(models.Manager):
+    def repayments(self, **kwargs):
```
noahsark769: I think you could probably call this `get_repayments`

revolv/payments/models.py
```diff
@@ -1,64 +1,235 @@
...
+from django.db.models import signals
+from django.dispatch import receiver
 from revolv.base.models import RevolvUserProfile
 from revolv.lib.utils import ImportProxy
 
-INSTRUMENT_PAYPAL = 'paypal'
-INSTRUMENT_CHECK = 'check'
-INSTRUMENT_REINVESTMENT = 'reinvestment'
-INSTRUMENT_REPAYMENT = 'repayment'
+PAYTYPE_PAYPAL = 'paypal'
```
noahsark769: These should really be class variables of PaymentType, haha

revolv/payments/models.py
```diff
@@ -1,64 +1,235 @@
...
+    total_left = instance.amount
+    pending_reinvestors = []
+    for user in RevolvUserProfile.objects.filter(reinvest_pool__gt=0.0):
+        total_left -= user.reinvest_pool
+        reinvest_amount = user.reinvest_pool + min(0.0, total_left)
+        pending_reinvestors.append((user, reinvest_amount))
+        if total_left <= 0.0:
+            break
+    for (user, amount) in pending_reinvestors:
+        reinvestment = Payment(user=user,
```
noahsark769: You could just use .create here :)

revolv/payments/models.py
```diff
@@ -1,64 +1,235 @@
...
+        if kwargs.get('project'):
+            queryset = queryset.filter(project=kwargs['project']).order_by('created_at')
+        if kwargs.get('admin_repayment'):
+            queryset = queryset.filter(admin_repayment=kwargs['admin_repayment']).order_by('created_at')
+        return queryset
+
+
+class Repayment(models.Model):
+    """
+    Abstraction for a repayment.
```
noahsark769: This should say how it's different from `AdminRepayment`

revolv/payments/models.py
```diff
@@ -78,50 +249,75 @@ def donations(self, user=None, project=None, queryset=None):
...
-        """
-        return self.payments(queryset=queryset).exclude(payment_instrument_type__name=INSTRUMENT_REINVESTMENT).exclude(payment_instrument_type__name=INSTRUMENT_REPAYMENT)
+        return AdminRepayment.objects.filter(revolvuserprofiles__in=user, projects__in=project).distinct()
 
-    def total_distinct_donors(self, queryset=None):
+    def total_distinct_organic_donors(self, queryset=None):
         """
-        :return: The total number of donors that have ever donated to a RE-volv
-            project. Useful for displaying stats on the homepage.
+        :return: The total number of organic donors that have ever donated to a
```
noahsark769: Would be good here to clarify what an "organic" donor is

revolv/payments/models.py
```diff
@@ -78,50 +249,75 @@ def donations(self, user=None, project=None, queryset=None):
...
-            queryset = self.all_donations()
+            queryset = self.donations()
         num_users = queryset.values("user").distinct().count()
         return num_users
 
 
 class Payment(models.Model):
     """
-        Abstraction indicating one particular payment.
+    Abstraction indicating one particular payment.
```
noahsark769: We should indicate here how a Payment is different from a Repayment

revolv/payments/tests/models/test_payment.py
```diff
@@ -1,54 +1,68 @@
...
-        self.assertEquals(Payment.objects.total_distinct_donors(), 2)
+        self.assertEquals(Payment.objects.total_distinct_organic_donors(), 2)
         self._create_payment(user3).save()
-        self.assertEquals(Payment.objects.total_distinct_donors(), 3)
+        self.assertEquals(Payment.objects.total_distinct_organic_donors(), 3)
 
     def test_payments(self):
         """Verify that we can create payments of any type and associate them to users."""
-        user1, user2 = RevolvUserProfile.factories.base.create_batch(2)
+        user1, user2, admin = RevolvUserProfile.factories.base.create_batch(3)
```
noahsark769: Ain't nobody got time for this admin variable

revolv/payments/tests/models/test_payment.py
```diff
@@ -56,46 +70,169 @@ def test_donations(self):
...
-        self._create_payment(user1, project2).save()
+        self._create_payment(user1, project=project2).save()
         self.assertEquals(Payment.objects.donations(user1, project2).count(), 1)
 
         self.assertEquals(Payment.objects.donations(user2).count(), 0)
 
-    def test_reinvestments(self):
-        """Test that we can bookkeep reinvestment information."""
+    def test_proportion_donated(self):
+        """Verify that repayment proportions are correct."""
```
noahsark769: This docstring could use some expanding

revolv/payments/tests/models/test_payment.py
```diff
@@ -56,46 +70,169 @@ def test_donations(self):
...
+        user1, user2, admin1 = RevolvUserProfile.factories.base.create_batch(3)
```
noahsark769: If this needs to be an admin, then you should at least do `RevolvUserProfile.factories.admin.create()` instead of `base`

revolv/payments/tests/models/test_payment.py
```diff
@@ -56,46 +70,169 @@ def test_donations(self):
...
 
-        self._create_payment(user1, project2).save()
+        self._create_payment(user1, project=project2).save()
         self.assertEquals(Payment.objects.donations(user1, project2).count(), 1)
 
         self.assertEquals(Payment.objects.donations(user2).count(), 0)
 
-    def test_reinvestments(self):
-        """Test that we can bookkeep reinvestment information."""
+    def test_proportion_donated(self):
```
noahsark769: This is a quality test, revolv++

revolv/payments/tests/models/test_payment.py
```diff
@@ -56,46 +70,169 @@ def test_donations(self):
...
+        self._create_admin_repayment(admin1, 100.00, project).save()
+        self._create_admin_reinvestment(admin1, 100.00, project).save()
+        self.assertEquals(Payment.objects.donations(project=project).count(), 2)
+        self.assertEquals(project.proportion_donated(user2), 10.0 / 40.0)
+        self.assertEquals(project.proportion_donated(user1), 30.0 / 40.0)
+
+    def test_repayment(self):
+        """Test repayment bookkeeping."""
+        admin1, admin2 = RevolvUserProfile.factories.base.create_batch(2)
+        admin1.make_administrator()
```
noahsark769: You don't need these lines, just do `RevolvUserProfile.factories.admin` instead

revolv/payments/tests/models/test_payment.py
```diff
@@ -56,46 +70,169 @@ def test_donations(self):
...
+        admin2.make_administrator()
+
+        user1, user2 = RevolvUserProfile.factories.base.create_batch(2)
+        project1, project2 = Project.factories.base.create_batch(2)
+
+        self._create_payment(user1, amount=10.00, project=project1).save()
+        self._create_payment(user2, amount=30.00, project=project1).save()
+        project1.complete_project()
```
noahsark769: Should we also assert that a project that is NOT completed cannot have admin repayments made on it?

revolv/payments/tests/models/test_payment.py
```diff
@@ -56,46 +70,169 @@ def test_donations(self):
...
         self.assertEquals(Payment.objects.reinvestments(user1).count(), 1)
         self.assertEquals(Payment.objects.reinvestments(user2).count(), 0)
 
-    def test_repayments(self):
-        """Test that we can bookkeep repayment information."""
-        user1, user2 = RevolvUserProfile.factories.base.create_batch(2)
-        project, project2 = Project.factories.base.create_batch(2)
+    def test_admin_reinvestment(self):
+        """
+        Test reinvestment on AdminReinvestment level.
```
noahsark769: More verbose docstring pls :)

revolv/payments/tests/models/test_payment.py
```diff
@@ -56,46 +70,169 @@ def test_donations(self):
...
+        """
+        user1, user2, admin1 = RevolvUserProfile.factories.base.create_batch(3)
```
noahsark769: Again, the admin should use the admin factory

revolv/payments/tests/models/test_payment_instrument_type.py
```diff
@@ -1,16 +1,12 @@
...
-from revolv.payments.models import (INSTRUMENT_CHECK, INSTRUMENT_PAYPAL,
-                                    INSTRUMENT_REINVESTMENT,
-                                    INSTRUMENT_REPAYMENT,
-                                    PaymentInstrumentType)
+from revolv.payments.models import (PAYTYPE_CHECK, PAYTYPE_PAYPAL,
+                                    PAYTYPE_REINVESTMENT, PaymentType)
 
 
-class PaymentInstrumentTypeTest(TestCase):
+class PaymentTypeTest(TestCase):
```
noahsark769: If we're gunna rename this class we should probably rename this test file also

revolv/payments/tests/models/test_payment.py
aykamko: done with newest diff

made a test too

revolv/payments/models.py
aykamko: done!

revolv/payments/models.py
aykamko: too lazy :cry:

revolv/payments/models.py
```diff
@@ -2,63 +2,231 @@
...
+
+    We need a single "contact point" representing a repayment so that admins
+    have the ability to "revoke" a repayment, if it was entered falsely.
+    Deleting an AdminRepayment also deletes any Repayment associated with it,
+    effectively erasing any trace of the repayment.
+
+    ::Signals::
+    pre_init
+        Make sure that related project is indeed complete, else throw a
+        ProjectNotCompleteException and to disallow instantiation of an invalid
```
noahsark769: I think the "and" should be delete from this sentence right?

revolv/payments/models.py
```diff
@@ -2,63 +2,231 @@
...
+        AdminRepayment.
+    post_save
+        When an AdminRepayment is saved, a Repayment is generated for all
+        donors to a project, each weighed by that donor's proportion of the
```
noahsark769: Nitpick: should be "weighted"

revolv/payments/models.py
```diff
@@ -2,63 +2,231 @@
...
+    """
+    Model representing a single, admin-controlled "contact point" for a
+    reinvestment to an ongoing RE-volv project.
+
+    RE-volv usually only reinvests into a project at its launch, but it is still
+    possible for an admin to put in a reinvestment at any time. Creating an
+    AdminReinvestment automatically pools money from users with a non-zero pool
+    of reinvestable money, prioritizing users that have a preference for the
+    Category of the project being reinvested into. (A user's pool of
+    reinvestable money consists of repayments to that user.) We generate a
```
noahsark769: This may be a bit of confusing wording, since even though a user has Repayment objects, a "repayment" is not made directly to a user, but rather from a community organization to RE-volv: we should say that the reinvestable pool of money consists of Repayment objects that were assigned to that user as part of a repayment.

I think this help distinguish the difference between a Repayment (the model) and a "repayment" (the action of there being new money repaid from a project added to the reinvestment pools of everyone who donated to that project), which is a worthy goal of this docstring

revolv/payments/models.py
```diff
@@ -2,63 +2,231 @@
...
+    We need a single "contact point" representing a reinvestment so that admins
+    have the ability to "revoke" a reinvestment, if it was entered falsely.
+    Deleting an AdminReinvestment also deletes any reinvestment Payments
+    associated with it, effectively erasing any trace of the reinvestment.
+
+    ::Signals::
+    pre_init
+        Raises a NotEnoughFundingException before __init__ if there are not
+        enough funds for this AdminReinvestment.
+    post_save
```
noahsark769: One thing I just thought of: what if .save() is called twice on an AdminRepayment? Technically I think that should never happen, but we should at least specify in the docstring that it should not, because if we save an AdminReinvestment twice then the distribution algorithm will be run twice.

Alternatively, we could only run the algorithm if it's the first time the model is being created (signal functions can accept a `created` kwarg for checking that I believe)

aykamko: Yeah, there's a `created` kwarg, I'll use that. The same should be done for `AdminReinvestment`.

revolv/payments/models.py
```diff
@@ -2,63 +2,231 @@
...
+    user is repayed for an AdminRepayment to a project. In other words, for user
+    U, project P, and AdminRepayment R, the amount of the Repayment will be:
+
+        ((U's donation to P) / (Total organic donations to P)) * R.amount
+
+    When a Repayment is generated/deleted, we automatically increment/decrement
+    the pool of reinvestable money in the related user.
+
+    Repayments are generated automatically when an AdminRepayment is created.
+    Repayments should never be created manually!
```
noahsark769: Think it's worth it to try to do something tricky to always prevent Repayments from being created manually? I don't think Django has support for this, but maybe you could do `Repayment(i_know_what_im_doing=True, ...)` or something like that haha...not sure

revolv/payments/models.py
```diff
@@ -2,63 +2,231 @@
...
+
+    ::Signals::
+    post_save
```
noahsark769: Same thing here as with AdminReinvestment, a Repayment should always only be saved once, right? Because if it's saved multiple times, it will increment the reinvest_pool multiple times. If you decide to just use the created kwarg like I suggested earlier, you could have a test to make sure the reinvest_pool is only incremented once when you have multiple times :)

revolv/payments/models.py
```diff
@@ -78,50 +246,64 @@ def donations(self, user=None, project=None, queryset=None):
...
 
-    def total_distinct_donors(self, queryset=None):
+    def total_distinct_organic_donors(self, queryset=None):
         """
-        :return: The total number of donors that have ever donated to a RE-volv
-            project. Useful for displaying stats on the homepage.
+        :return:
+            The total number of organic donors that have ever donated to a
+            RE-volv project. Useful for displaying stats on the homepage.
+            Organic donors are donors that have made a non-reinvestment type
```
noahsark769: Organic donors here DON'T include donations via check which have a user=None, right? I feel like that would be good to note here

revolv/payments/models.py
```diff
@@ -78,50 +246,64 @@ def donations(self, user=None, project=None, queryset=None):
...
         num_users = queryset.values("user").distinct().count()
         return num_users
 
 
 class Payment(models.Model):
     """
-        Abstraction indicating one particular payment.
+    Abstraction indicating one particular payment.
+
+    A Payment represents only money flowing from a user to a project, not the
```
noahsark769: +1 quality explanation!!

revolv/payments/signals.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@receiver(signals.pre_init, sender=AdminRepayment)
+def pre_init_admin_repayment(**kwargs):
+    """
+    Make sure that related project is indeed complete, else throw a
+    ProjectNotCompleteException and to disallow instantiation of an invalid
+    AdminRepayment.
+    """
+    init_kwargs = kwargs.get('kwargs')
```
noahsark769: It's not immediately obvious to me why not having a `kwargs` kwarg would mean that there's not enough funding, could we have a comment explaining that or something?

revolv/payments/signals.py
```diff
@@ -0,0 +1,146 @@
...
+    invest_amount = init_kwargs.get('amount') or 0.0
+
+    global_repay_amount = AdminRepayment.objects.aggregate(
+        Sum('amount')
+    )['amount__sum'] or 0.0
+    global_reinvest_amount = AdminReinvestment.objects.aggregate(
+        Sum('amount')
+    )['amount__sum'] or 0.0
+    global_reinvest_pool = global_repay_amount - global_reinvest_amount
+    if global_reinvest_pool - invest_amount < 0.0:
```
noahsark769: Should be <= ? Since there's not enough funding if there are 0.0 repayments right?

revolv/payments/signals.py
```diff
@@ -0,0 +1,146 @@
...
+    total_left = instance.amount
+    pending_reinvestors = []
+    for user in RevolvUserProfile.objects.filter(reinvest_pool__gt=0.0):
+        total_left -= user.reinvest_pool
+        reinvest_amount = user.reinvest_pool + min(0.0, total_left)
+        pending_reinvestors.append((user, reinvest_amount))
+        if total_left <= 0.0:
+            break
+    for (user, amount) in pending_reinvestors:
+        reinvestment = Payment(user=user,
```
noahsark769: I feel like a comment here reminding the reader that the `user.reinvest_pool` will be decremented by the post_save signal on the Payment would be helpful

revolv/payments/signals.py
```diff
@@ -0,0 +1,146 @@
...
+
+@receiver(signals.post_save, sender=Payment)
+def post_save_payment(**kwargs):
+    """
+    If the payment is organic, we add this payment's user as a donor to the
+    related project. If the payment is a reinvestment, we decrement the
+    reinvest_pool in the related user.
+    """
+    instance = kwargs.get('instance')
+    if instance.payment_type == PaymentType.objects.get_paypal():
```
noahsark769: Supposedly we should compare whether the user == the entrant, right? Because when we put in paying with paypal directly instead of paying with CC processed via paypal, we'll have to change this (because both types will be considered an organic donation).

Even better would be to write a `Payment.isOrganic` method haha

revolv/payments/tests/models/test_payment.py
```diff
@@ -1,101 +1,270 @@
...
         if project is None:
             project = Project.factories.base.create()
-        if instrument_type is None:
-            instrument_type = PaymentInstrumentType.objects.get_paypal()
-        return Payment(amount=10.00, user=user, entrant=user, payment_instrument_type=instrument_type, project=project)
+        if payment_type is None:
+            payment_type = PaymentType.objects.get_paypal()
+        return Payment(amount=amount, user=user, entrant=user, payment_type=payment_type, project=project)
+
+    def _create_admin_repayment(self, admin, amount=10.00, project=None):
```
noahsark769: Didn't we decide we didn't need these functions?

revolv/payments/tests/models/test_payment.py
```diff
@@ -1,101 +1,270 @@
...
+
+        project.complete_project()  # must complete project before repayment
+
+        self._create_admin_repayment(admin1, 100.00, project).save()
+        self._create_admin_reinvestment(admin1, 100.00, project).save()
+        self.assertEquals(Payment.objects.donations(project=project).count(), 2)
+        self.assertEquals(project.proportion_donated(user2), 10.0 / 40.0)
+        self.assertEquals(project.proportion_donated(user1), 30.0 / 40.0)
+
+    def test_bad_repayment(self):
```
noahsark769: Great test, shouldn't we have one for reinvestments also that asserts that NotEnoughFundingException is thrown in the right places?

revolv/project/models.py
```diff
@@ -302,17 +302,42 @@ def update_categories(self, category_list):
     def get_absolute_url(self):
         return reverse("project:view", kwargs={"pk": str(self.pk)})
 
+    def get_organic_donations(self):
+        return self.payment_set.filter(payment_type=PaymentType.objects.get_paypal())
+
+    def proportion_donated(self, user):
```
noahsark769: Might be good to have a `:return:` here mentioning that it returns a float between 0 and 1 inclusive

noahsark769:
Alright haha, dang I made a lot of comments. Most stuff is just docstring/comment nitpicks though, but I think they're important since this is probably the part of the project where documentation matters the most.

Also, I'm pretty sure we should be doing most of these post_save signals only if the model is being created for the first time (this is also an opportunity for most tests wooo)

noahsark769:
Almost ready to ship though!!

revolv/payments/signals.py
aykamko: Uh, pretty sure this is right. It's checking to see that we won't dip below `$0` if we make this reinvestment.

noahsark769: Oh lol I guess you're right haha, it would also only apply if a
reinvestment had an amount of 0.0

aykamko:
Ok, most nits fixed except:
1. I kept the `_create_payment` helper test functions because refactoring proved to be a huge pain. We have *a lot* of tests now lol
2. I think preventing individual RepaymentFragments from being created would be annoying, so I'm just gonna hope that it doesnt happen

revolv/payments/signals.py
noahsark769: Oh yeah wow nevermind, I'm dumb. It will be equal to zero any time we reinvest the total pool, which will be every time, and we want that to work

revolv/payments/models.py
```diff
@@ -2,63 +2,253 @@
...
-           part the revolving funding model - the money originally came from an organic
-           donation to an already completed project, and then the community repayed
-           revolv the money that is represented in the Payment that has this PaymentInstrumentType.
+            queryset = super(PaymentTypeManager, self).get_queryset()
+        return queryset.get(name=PaymentType._REINVESTMENT)
+
+
+class PaymentType(models.Model):
     """
+    Abstraction for a payment. A payment type is defined as anything that can
```
noahsark769: Random typo here (should be abstraction for type of payment, not for payment), I'm pretty sure this is the only issue right now though, so if it is I'll just push the one word fix haha

noahsark769:
Ahhhhh okay I hate to say this, but I think we need just one more thing: tests for how the "check" payment type behaves (i.e. that it is not included in organic donations and thus not included in calculating proportional donation amounts and thus not included in reinvestments).

Also there's the typo that I mentioned, which I guess you can do now since the check stuff will involve writing a couple of more tests (SORRY, testing is really important for this part of the app :( ). On the plus side, we have 60+ tests now and they all pass which is AWESOME!

noahsark769:
@aykamko let me know when you push up the new tests :)

noahsark769:
Awww yeah revolv++ finally! Thanks a lot man!

PR #204, opened 2015-03-17
Implement donation levels for projects

### LINKED ISSUE #191: Implement "donation levels"
Similar to indiegogo and kickstarter, we should have itemized descriptions of how different dollar amounts can contribute to a project.

1. They should be on the project page
2. They should be editable on a per project bases (one to many project to project level)
3. They should include a dollar amount and a description (optional). The UI should show stats, but those can be calculated via percentages of the donation amount over the total goal of the project times the actual project stats.

erictshen:
Fixes #191 
revolv++

Lots of inspiration from http://www.lab305.com/news/2012/jul/19/django-inline-formset-underscore/
http://jacobian.org/writing/dynamic-form-generation/
http://kevindias.com/writing/django-class-based-views-multiple-inline-formsets/
https://docs.djangoproject.com/en/1.7/topics/forms/modelforms/#inline-formsets

### DIFF
--- a/revolv/project/models.py
+++ b/revolv/project/models.py
@@ -398,7 +398,7 @@ class Project(models.Model):
 
     @property
     def donation_levels(self):
-        return DonationLevel.objects.filter(project_id=self.id).values()
+        return self.donationlevel_set.all()
 
 
 class Category(models.Model):
--- a/revolv/project/models.py
+++ b/revolv/project/models.py
@@ -396,6 +396,10 @@ class Project(models.Model):
     def categories(self):
         return [category.title for category in self.category_set.all()]
 
+    @property
+    def donation_levels(self):
+        return DonationLevel.objects.filter(project_id=self.id).values()
+
 
 class Category(models.Model):
     """
--- a/revolv/project/models.py
+++ b/revolv/project/models.py
@@ -3,6 +3,7 @@ from itertools import chain
 
 from django.core.urlresolvers import reverse
 from django.db import models
+
 from imagekit.models import ImageSpecField, ProcessedImageField
 from imagekit.processors import ResizeToFill
 from revolv.base.models import RevolvUserProfile
@@ -413,3 +414,12 @@ class Category(models.Model):
 
     def __unicode__(self):
         return self.title
+
+
+class DonationLevel(models.Model):
+    project = models.ForeignKey(Project)
+    description = models.CharField(max_length=200)
+    amount = models.DecimalField(
+        max_digits=15,
+        decimal_places=2
+    )
--- a/revolv/project/models.py
+++ b/revolv/project/models.py
@@ -398,6 +398,10 @@ class Project(models.Model):
 
 
 class Category(models.Model):
+    """
+    Categories that a project is associated with. Categories are predefined,
+    and as of now, loaded through fixtures.
+    """
     HEALTH = 'Health'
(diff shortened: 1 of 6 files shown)


erictshen:
screen shot 2015-03-16 at 9 01 51 pm
screen shot 2015-03-16 at 9 02 22 pm
screen shot 2015-03-16 at 9 02 49 pm

revolv/project/models.py
```diff
@@ -395,8 +396,16 @@ def is_completed(self):
     def categories(self):
         return [category.title for category in self.category_set.all()]
 
+    @property
+    def donation_levels(self):
+        return DonationLevel.objects.filter(project_id=self.id).values()
```
noahsark769: Woah, this works? project_id isn't a field of DonationLevel. At any rate, it would be better to just return self.donationlevel_set, correct?

revolv/project/views.py
```diff
@@ -62,15 +77,28 @@ def get_success_url(self):
         messages.success(self.request, 'Project details updated')
         return reverse('project:view', kwargs={'pk': self.get_object().id})
 
+    # validates project, formset of donation levels, and adds categories as well
     def form_valid(self, form):
-        project = self.get_object()
-        project.update_categories(form.cleaned_data['categories_select'])
+        context = self.get_context_data()
```
noahsark769: Can we implement like a `get_formset` method for this? It's fairly weird to be calling `get_context_data` in a POST request

noahsark769:
Just a couple of things (I commented on them). Also, can we at least format the perks in some sort of way other than the default python stringification? It's super weird to have that JSON like syntax in the template, even if it will be rewritten soon

revolv/project/views.py
erictshen: Will look into this mid break (like Wednesday-ish). Unfortunately, there might not be a way to do this without calling get_context_data (though I can define a helper method)

noahsark769: I think helper method is the way to go

noahsark769:
@erictshen let me know when you can take a look at the get_context_data stuff

erictshen:
Added helper methods and renamed some of the objects and methods to make it more understandable. Let me know if there's anything else to change!

erictshen:
I see. Makes sense, and it works!

noahsark769:
Okay wait, so now we have the exact same method in two classes. Can we extract that into a `DonationLevelFormsetMixin` (with appropriate docstring for what the one method it provides does).

Also, for categories display on the project page, you changed it to not have the JSON formatting right?

erictshen:
Updated pull request to use DonationLevelFormSetMixin

Also fixed project page to show categories not in JSON

screen shot 2015-03-25 at 5 48 43 pm

revolv/project/views.py
```diff
@@ -15,7 +16,21 @@
 from revolv.project.models import Category, Project
 
 
-class CreateProjectView(CreateView):
+class DonationLevelFormSetMixin(object):
+    """
+    Mixin that gets the ProjectDonationLeveLFormSet for a page, specifically
+    the Create Project and Update Project page.
+    """
+    # checks if the request is a POST, and populates the formset with current object as the instance
```
noahsark769: This should be a docstring of the method :)

noahsark769:
Looks great!

PR #212, opened 2015-03-31
Sameera project page styles

sameerav:
List of things that still don't work:

1. [FIXED] The iframe still isn't clickable
2. [FIXED] The top donate button only somehow has half the button being clickable (weird anchor tag positioning). (FIGURED THIS OUT! this is because the modal is covering it, so as soon as the modal thing is fixed this should be good).
3. The donate modal doesn't have a top attribute of 0 when the page is scrolled down, so when we click the donate button at the bottom of the page, it doesn't work.
4. [FIXED] There's a tiny little margin at the top of the project page image.
5. [FIXED] [EDIT by Noah] Line breaks should be preserved in project updates


noahsark769:
Can you try to fix part 4 before we merge this pull request? I'm going to go through and style check everything also, but I feel like that tiny margin shouldn't be put in because it's a but that isn't related to anything else (as far as I know)

revolv/project/forms.py
```diff
@@ -66,15 +66,37 @@ class Meta:
         fields = ()
 
 
-class PostFundingUpdateForm(forms.ModelForm):
+# class PostFundingUpdateForm(forms.ModelForm):
```
noahsark769: Instead of commenting this out, just delete it :)

revolv/project/forms.py
```diff
@@ -66,15 +66,37 @@ class Meta:
...
+        model = ProjectUpdate
+        fields = (
+            'update_text',
+        )
+
+class EditProjectUpdateForm(forms.ModelForm):
+    """ 
+    A form used to edit updates about a project
+    """
+    #SET SOME KIND OF DEFAULT VALUE
```
noahsark769: Delete :)

revolv/project/models.py
```diff
@@ -245,6 +244,9 @@ class Project(models.Model):
         default=0.0,
         help_text='The internal rate of return for this project.'
     )
+    
+    #TAKE NOTICE AND GET RID OF THIS LATER
```
noahsark769: We should resolve the error that it creates :D

revolv/project/models.py
```diff
@@ -336,6 +338,17 @@ def amount_donated_organically(self):
         )["amount__sum"] or 0.0
 
     @property
+    def location_street(self):
+        return self.location.split(',')[0]
+
+    @property
+    def location_city_state_zip(self):
```
noahsark769: Are there locations in which there might not be a city, or a state, or a zip, or a street? I feel like for street, there probably are. We need to make sure these functions dont error in those cases

noahsark769: Also I think this function could use a docstring

revolv/project/models.py
```diff
@@ -384,6 +397,15 @@ def partial_completeness(self):
         ratio = self.amount_donated / float(self.funding_goal)
         return min(ratio, 1.0)
 
+    @property
+    def percent_complete(self):
+        """
+        :return: an int between 0 and 100, representing the completeness of this
```
noahsark769: We should note in this docstring that the percent will be a floor of the actual percent (e.g. 97.5 => 97)

revolv/project/models.py
```diff
@@ -426,6 +448,43 @@ def categories(self):
         return [category.title for category in self.category_set.all()]
 
     @property
+    def updates(self):
+        return self.update_set.all()
+
+    def add_update(self, text):
```
noahsark769: This could use a short docstring I think

revolv/project/models.py
```diff
@@ -426,6 +448,43 @@ def categories(self):
...
+        update = ProjectUpdate(update_text=text, project=self)
+        update.save()
+
+class ProjectUpdate(models.Model):
+    update_text = models.TextField(
```
noahsark769: Why not just `text`?

revolv/project/models.py
```diff
@@ -426,6 +448,43 @@ def categories(self):
...
+        help_text = "What time was your update created?",
+        auto_now_add = True
+    )
+    
+    project = models.ForeignKey(
+        Project,
+        related_name="update"
+    )
+
+    def set_update_text(self, text):
```
noahsark769: Why do we need this function? Can't we just set the update text...?

revolv/project/models.py
```diff
@@ -426,6 +448,43 @@ def categories(self):
...
+        self.update_text = text
+
+    @property
+    def month_word(self):
```
noahsark769: If this is for rendering in the template, you should use [django.contrib.humanize](https://docs.djangoproject.com/en/1.8/ref/contrib/humanize/) instead

noahsark769: Oh oops, actually humanize doesn't do that. But [django's date filter](https://docs.djangoproject.com/en/dev/ref/templates/builtins/#date) does

revolv/project/tests/test_integration.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class PostProjectUpdatesTest(TestUserMixin, UserTestingMixin, TestCase):
+
+    def helper_test_create_update_ambassador_administrator(self):
+        """
+        Tests that administrators and ambassadors can create project updates.
+        """
+        response = self.send_test_user_login_request()
+        self.assertUserAuthed(response)
```
noahsark769: If you're doing asserts in this helper function, then its name should start with `assert`. For this function, looks like `assert_user_can_create_project_update` would be a good name

revolv/project/tests/test_integration.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        Tests that donors can't create project updates.
+        """
+        self.test_profile.make_donor()
+        response = self.send_test_user_login_request()
+        self.assertUserAuthed(response)
+        project = Project.factories.base.create()
+        response = self.client.post('/project/%d/update' % project.pk, {'update_text': 'This is an update'})
+        
+        #the deny_access method is called in the view from the userdatamixin, which returns a redirect to the homepage
```
noahsark769: Space after # in comments please :)

revolv/project/tests/test_integration.py
```diff
@@ -0,0 +1,185 @@
...
+        self.assertEqual(response.status_code, 302)
+        
+class EditProjectUpdatesTest(TestUserMixin, UserTestingMixin, TestCase):
+
+    def _make_update(self, project, text):
+        """
+        Makes an update to of given text to the given project.
```
noahsark769: Typo here

revolv/project/tests/test_integration.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        response = self.send_test_user_login_request()
+        self.assertUserAuthed(response)
+        #response = self.client.post('/project/%d/update' % project.pk, {'update_text': text})
+        ProjectUpdate(project = project, update_text = text).save()
+        new_updates = ProjectUpdate.objects.filter(update_text = text)
+        return new_updates[0]
+
+    def _helper_test_edit_update(self, donor = False):
```
noahsark769: Again, as above, this needs a more descriptive name

revolv/project/tests/test_integration.py
```diff
@@ -0,0 +1,185 @@
...
+
+        old_update_text_updates = ProjectUpdate.objects.filter(update_text="This update has not been changed.")
+        new_updates = ProjectUpdate.objects.filter(update_text="This update has been changed.")
+
+        if not donor:
+            self.assertEqual(len(old_update_text_updates), 0)
+            self.assertEqual(update.project, new_updates[0].project)
+            self.assertEqual(new_updates[0].update_text, "This update has been changed.")
+        else:
+            #the donor should be denied access and redirected
```
noahsark769: Space after #

revolv/project/tests/test_integration.py
```diff
@@ -0,0 +1,185 @@
...
+            self.assertEqual(response.status_code, 302)
+
+    def test_edit_update_administrator(self):
+        """
+        Tests that a change is made to an update for an administrator.
+        """
+        self.test_profile.make_administrator()
+        self._helper_test_edit_update()
```
noahsark769: Might as well pass in the False kwarg here for readability purposes, I think. `donor=False` is a lot more descriptive than just nothing

revolv/project/tests/test_models.py
```diff
@@ -11,6 +12,31 @@
...
+
+        update1.save()
+        update2.save()
+        self.assertEqual(len(ProjectUpdate.objects.all()),2)
+
+    def test_add_update(self):
+        project = Project.factories.base.create()
+        project.add_update('Another sample update')
+        update3 = ProjectUpdate.objects.get(update_text='Another sample update')
+        self.assertEqual(project.id, update3.project_id)
```
noahsark769: You should compare `project` and `update3.project` here instead. Also, why is it called update3? why not just update or test_update?

revolv/project/tests/test_models.py
```diff
@@ -217,44 +243,26 @@ def test_update_category_multiple_projects(self):
...
+        # self.assertEqual(resp.status_code, 200)
+        
         # note: if the link makes a modal appear, it will be skipped and the
         # test will fail because it couldn't find the link - this is what
         # we want to happen in this case, but we may have to change this if
         # we want a login modal to appear instead.
-        resp = (resp.click(linkid="donate-button")).maybe_follow()
-        self.assertTemplateUsed(resp, "base/sign_in.html")
+
+        # commented out because this test is deprecated and a login modal
```
noahsark769: wait what we have a login modal?

revolv/project/urls.py
```diff
@@ -11,5 +12,6 @@
     url(r'^(?P<pk>\d+)/review$', is_ambassador(ReviewProjectView.as_view()), name='review'),
     url(r'^(?P<pk>\d+)/$', ProjectView.as_view(), name='view'),
     url(r'^(?P<pk>\d+)/donation/submit$', SubmitDonationView.as_view(), name="donation_submit"),
-    url(r'^(?P<pk>\d+)/update$', is_ambassador(PostFundingUpdateView.as_view()), name='update'),
+    url(r'^(?P<pk>\d+)/update$', is_ambassador(PostProjectUpdateView.as_view()), name='update'),
+	  url(r'^editupdate/(?P<pk>\d+)$', is_ambassador(EditProjectUpdateView.as_view()), name='edit_project_update'),
```
noahsark769: This should be the same level of indentation?

revolv/project/views.py
```diff
@@ -155,20 +157,64 @@ def get_context_data(self, **kwargs):
...
+class PostProjectUpdateView(UserDataMixin, UpdateView):
+    model = Project
+    template_name = 'project/post_project_update.html'
+    form_class = forms.PostProjectUpdateForm
 
-class PostFundingUpdateView(UpdateView):
-    """
-    The view to send out post funding updates about a project after it has completed.
+    def dispatch(self, request, *args, **kwargs):
+        self.user = request.user
```
noahsark769: You should already have `self.user` and `self.is_authenticated` from `UserDataMixin`

revolv/project/views.py
```diff
@@ -155,20 +157,64 @@ def get_context_data(self, **kwargs):
...
-        return reverse('project:view', kwargs={'pk': self.get_object().id})
+        return reverse('project:review', kwargs={'pk': self.get_object().id})
+
+    def form_valid(self, form):
+        text = form.cleaned_data['update_text']
+        project = self.get_object()
+        project.add_update(text)
+        return super(PostProjectUpdateView, self).form_valid(form)
+
+    def form_invalid(self, form):
```
noahsark769: We need to redirect back to the page with an error message here, lol. Also: we don't allow empty updates, right? Do we have validation for that on the form?

revolv/project/views.py
```diff
@@ -155,20 +157,64 @@ def get_context_data(self, **kwargs):
...
+        print("oh no", form)
+
+class EditProjectUpdateView(UserDataMixin, UpdateView):
+    
+    model = ProjectUpdate
+
+    #STILL HAVE TO MAKE THIS TEMPLATE AND FORM
```
noahsark769: Delete :)

revolv/project/views.py
noahsark769: Actually you should have almost all this stuff haha

revolv/project/views.py
```diff
@@ -155,20 +157,64 @@ def get_context_data(self, **kwargs):
...
+    template_name = 'project/edit_project_update.html'
+    form_class = forms.EditProjectUpdateForm
+
+    def dispatch(self, request, *args, **kwargs):
+        self.user = request.user
```
noahsark769: Again, the next five lines are already in UserDataMixin

revolv/project/views.py
```diff
@@ -155,20 +157,64 @@ def get_context_data(self, **kwargs):
...
+        return reverse('project:review', kwargs={'pk': self.get_object().project_id})
+
+    def form_valid(self, form):
+        text = form.cleaned_data['update_text']
+        update = self.get_object()
+        update.set_update_text(text)
+        return super(EditProjectUpdateView, self).form_valid(form)
+
+    def form_invalid(self, form):
+        print("oh no", form)
```
noahsark769: Need to do an actual thing -_-

revolv/project/views.py
```diff
@@ -183,6 +229,8 @@ class ProjectView(UserDataMixin, DetailView):
     def get_context_data(self, **kwargs):
         context = super(ProjectView, self).get_context_data(**kwargs)
         context['GOOGLEMAPS_API_KEY'] = settings.GOOGLEMAPS_API_KEY
+        context['updates'] = self.get_object().update.all()[::-1]
```
noahsark769: Probably better to do an `order_by` instead of a list reverse here?

revolv/project/views.py
```diff
@@ -183,6 +229,8 @@ class ProjectView(UserDataMixin, DetailView):
...
+        context['donor_count'] = len(self.get_object().donors.all())
```
noahsark769: You can go `.donors.count()` :)

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,68 @@
+$(document).ready(function () {
+	drawPercentagePartialCircle()	
+
```
noahsark769: This line can be deleted

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,68 @@
...
+    var partialCompleteness = $(".storage-div").attr("id");
+
+    //use this for test purposes so you can see the progress bar
+    // var partialCompleteness = 0.7;
+    
+    if (partialCompleteness <= 0.01 && partialCompleteness !== 0.0) partialCompleteness = 0.02;
+    drawD3PartialTriangle(partialGrouping, ["project-badge-line"], radius, padding, partialCompleteness);
+}
+
+var drawD3PartialTriangle = function (destination, classes, radius, padding, partial) {
```
noahsark769: This function is now duplicated from home.js, so we should extract it into a utility file so as to not repeat code. Also, the name draw3dPartialTriangle is kind of a misnomer, since it actually draws a circle...sorry about that, that's my fault lol. Do you mind renaming it and also writing a docstring?

noahsark769: (Using [jsDoc](http://usejsdoc.org/) syntax)

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,68 @@
...
+});
+
+var drawPercentagePartialCircle = function() {
+    var radius = 100,
+        padding = 10,
+        radians = 2 * Math.PI;
+
+    if ($(document).width() > 200 && $(document).width() < 800) {
+        radius = 48;
+        //radius = $(document).width()*0.16;
```
noahsark769: Deletable

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,68 @@
...
+        .attr("width", dimension)
+        .attr("height", dimension)
+        .append("g");
+
+    var circleGrouping = svg.append("g").attr("class", "project-badge-circle-grouping");
+    var partialGrouping = svg.append("g").attr("class", "project-badge-partial-grouping");
+
+    drawD3PartialTriangle(circleGrouping, ["project-badge-circle"], radius, padding, 1);
+
+    console.log(window.project)
```
noahsark769: This shouldn't be here, doesn't jsHint catch this?

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,68 @@
...
+        radians = 2 * Math.PI;
+
+    if ($(document).width() > 200 && $(document).width() < 800) {
+        radius = 48;
+        //radius = $(document).width()*0.16;
+    }
+
+    var dimension = (2 * radius) + (2 * padding);
+
+    var svg = d3.select(".internal-graphics-container")
```
noahsark769: I think that the target class should be an argument to this function instead of a hardcoded string, so that if we change the class name in the html we don't have to change it here also

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,68 @@
...
+        .append("g");
+
+    var circleGrouping = svg.append("g").attr("class", "project-badge-circle-grouping");
+    var partialGrouping = svg.append("g").attr("class", "project-badge-partial-grouping");
+
+    drawD3PartialTriangle(circleGrouping, ["project-badge-circle"], radius, padding, 1);
+
+    console.log(window.project)
+    
+    //this line will get actual partial completeness
```
noahsark769: One space after the // please :)

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,68 @@
...
+    var partialCompleteness = $(".storage-div").attr("id");
+
+    //use this for test purposes so you can see the progress bar
+    // var partialCompleteness = 0.7;
+    
+    if (partialCompleteness <= 0.01 && partialCompleteness !== 0.0) partialCompleteness = 0.02;
```
noahsark769: Use braces for if statements, every time, all the time :D :D :D

noahsark769: That may be my fault also, sorry :(

revolv/static/mixins.scss
```diff
@@ -31,3 +31,169 @@
     -webkit-transform: $transforms;
     transform: $transforms;
 }
+
+.revolv-reveal-modal-table {
```
noahsark769: I feel like this file should be for `@include`s and such and that the common modal and check styles should go in a `modal.scss`. What do you think?

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
     }
 }
 
+body {
+    font-family: Lato;
```
noahsark769: This needs a fallback font: this font family is defined in revolv-settings.scss

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+}
+
+.row {
```
noahsark769: You shouldn't add styles like this to a foundation class, because they'll get added to ALL rows on the page (or maybe that's what you want)?

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+    margin-right: 0px;
+    margin-left: 0px;
+    max-width: 100rem;
```
noahsark769: Do you need this max-width?

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+    &.dashboard {
+        height: 600px;
+    }
+
+    &.categories {
+        max-width: 100%;
+    }
+}
+
+.header-img-div {
```
noahsark769: having a class of `header-img-div` is kind of weird since technically the element might not be a div at all. How about `header-img-container` instead?

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+    width: 100%;
+    height: 500px;
+    position: absolute;
```
noahsark769: Why does this need to be position: absolute? You're not setting bottom, top, left, or right

noahsark769: Oh, it's for the child! In this case, this element should be position: relative instead.

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+    .overlay {
+        width: 100%;
+        height: 100%;
+        top: 0;
+        left: 0;
+        background: rgba(0, 0, 0, .5);
+        position: absolute;
+        
+        .donate-button-header {
+            border-color: white;
```
noahsark769: You can just do border: 2px solid white

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            border-width: 2px;
+            border-style: solid;
+            width: 15%;
+            height: 10%;
+            top: 410px;
```
noahsark769: Shouldn't this be offset some amount from the bottom rather than from the top?

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            left: 80%;
```
noahsark769: Same thing, should probably be offset from the right?

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            z-index: 100;
+            position: absolute;
+            transition: all 0.3s;
+
+            .donate-link {
+                width: 100%;
+                height: 100%;
+                display: inline-block;
+                
+                p {
```
noahsark769: If you actually look in the web inspector, this element overflows the button on the bottom. I think it's because it had a line-height that is set elsewhere maybe?

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            }
+        }
+
+        .project-title {
+            color: white;
+            font-size: 2rem;
+            font-weight: 800;
+            top: 400px;
+            left: 5%;
+            position: absolute;
```
noahsark769: Just a style thing, the position: absolute definition is more readable if it's before the top and left definitions

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+
+        }
+    }
+}
+
+.dashboard {
+    height: 400px;
+    z-index: 1000;
+    position: absolute;
+    top: 300px;
```
noahsark769: I think all these position offsets should at least be variables to make this code more semantic

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+        height: 100%;
+        width: 100%;
+
+        .top-bar {
+            background-color: $revolv-blue;
+            width: 100%;
+            height: 40px;
+            position: absolute;
+            top: 50%;
+            display: inline-block;
```
noahsark769: You don't need to make it inline-block if its absolute, since flow isn't taken into account for absolutely positioned elements anyway

noahsark769: (Same thing for bottom bar)

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            margin: 0;
+
+            .container {
```
noahsark769: This could use a more semantic name, I think

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+
+        .bottom-bar {
+            background-color: #064f7e;
+            width: 100%;
+            height: 40px;
+            position: absolute;
+            top: 60%;
+            display: inline-block;
+
+            .container {
```
noahsark769: Looks like a lot of styles are repeated between top and bottom bar. Can we make their definitions more DRY? (DRY = Don't Repeat Yourself)

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            }
+        }
+
+        .svg-graphics-container {
+            height: 100%;
+            width: 100%;
+            position: absolute;
+        }
+
+        .internal-graphics-container {
```
noahsark769: Why do you need both `svg-graphics-container` and `internal-graphics-container`?

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            top: 50%;
+            -webkit-transform: translate(-50%, -50%);
+            -ms-transform: translate(-50%, -50%);
+            transform: translate(-50%, -50%);
+            height: 220px;
+            width: 220px;
+
+            .percentage-text {
+                font-size: 3rem;
+                color: #064f7e;
```
noahsark769: This should be a revolv-settings setting

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+          stroke: white;
+        }
+
+        .text-container {
+            position: absolute;
+            left: 50%;
+            top: 50%;
+            -webkit-transform: translate(-50%, -50%);
+            -ms-transform: translate(-50%, -50%);
+            transform: translate(-50%, -50%);
```
noahsark769: Why do you need all this translation? They won't work in IE 8 ([http://caniuse.com/#feat=transforms2d](http://caniuse.com/#feat=transforms2d))

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+                left: 50%;
+                top: 40%;
+                -webkit-transform: translate(-50%, -50%);
+                -ms-transform: translate(-50%, -50%);
+                transform: translate(-50%, -50%);
+            }
+
+            .funded {
+                font-size: 1.5rem;
+                color: #064f7e;
```
noahsark769: If you have this same color on both funded and percentage-text, it should probably be on the parent instead. I also don't think these elements need to be position: absolute either right?

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+        }
+
+        .category-holder {
+            display: inline-block;
+            border-color: #888888;
+            border-style: solid;
+            border-width: 1px;
+            width: 15%;
+            height: 30px;
+            margin-top: 10px;
```
noahsark769: You can define these three margins inline with `margin:`

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            margin-right: 5px;
+            margin-left: 5px;
+
+            .category-label {
+                color: #888888;
+                text-align: center;
+                line-height: 30px;
```
noahsark769: This should be 28px to account for the border of the boxes

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            }
+        }
+    }
+
+    .about-text {
```
noahsark769: I think in the mockup we had the text here as dark gray rather than black? Black is fairly jarring on this page I think

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+        width: 100%;
+        margin-bottom: 40px;
+        display: inline-block;
+        position: relative;
+
+        .date-container {
+            height: 50px;
+            width: 100%;
+            text-align: center;
+            color: #064f7e;
```
noahsark769: Again, revolv-settings

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+            font-weight: 300;
+            line-height: .5;
```
noahsark769: Nitpick, but 0.5 is nicer than just .5

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+
+            .date-year {
+                font-size: 4.5rem;
+                margin-bottom: 0rem;
+                line-height: 1;
```
noahsark769: Can you put units on this? haha

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+        }
+    }
+}
+
+.map {
+    height: 400px;
+    margin-top: 40px;
+    margin-bottom: 40px;
+
+    #map-canvas {
```
noahsark769: In general don't target ids :)

revolv/static/project.scss
```diff
@@ -43,6 +43,439 @@
...
+    }
+
+
+    .overlay {
+        background: rgba(0, 0, 0, .5);
+        float: left;
+        height: 400px;
+        width: 40%;
+        position: absolute;
+        margin-top: -400px;
```
noahsark769: I think this div should be position: absolute with height: 100%. That way you don't need the hacky negative margin

revolv/static/project.scss
```diff
@@ -77,3 +510,207 @@
     height: 300px;
     margin: 0 20px 20px 20px;
 }
+
+
+@media screen and (max-width: 1050px) {
+    .project-updates {
+
+        .small-12 {
```
noahsark769: You should definitely target this div in a different way, if we change it to take up a smaller number of columns, css will break

revolv/static/project.scss
```diff
@@ -77,3 +510,207 @@
     height: 300px;
     margin: 0 20px 20px 20px;
 }
+
+
+@media screen and (max-width: 1050px) {
```
noahsark769: For these media queries, you can use foundation builtin stuff, like `@media #{$small-only} and (orientation: portrait) {` instead of `@media screen and (max-width: 1050px) {`. Check out home.scss for examples

revolv/static/project.scss
```diff
@@ -77,3 +510,207 @@
...
+                height: 40px;
+                top: 50%;
+
+                .container {
+
+                    .amount-donated {
+                        width: 40%;
+                        line-height: 40px;
+                        font-size: 0.7rem;
+                        margin: 0px 0px 0px 10px;
```
noahsark769: When you have 0, you don't need to specify units: `0 0 0 10px`

revolv/static/project.scss
```diff
@@ -77,3 +510,207 @@
...
+                    .num-donors {
+                        line-height: 40px;
+                        font-size: 0.7rem;
+                        margin: 0px 10px 0px 0px;
+                    }
+                }
+            }
+
+            .svg-graphics-container {
+                width:0%;
```
noahsark769: You need spaces after the colons here

revolv/static/project.scss
```diff
@@ -77,3 +510,207 @@
...
+                height:0%;
+                padding:65px;
+                border-radius:100%;
+                top: 215px;
+                left: 50%;
+                -webkit-transform: translateX(-50%);
```
noahsark769: Are these transforms trying to center the div horizontally? If so, you should just make it an inline element and set text-align: center on the parent. Also, you don't specify the regular `transform`?

revolv/static/project.scss
noahsark769: Also, no need for % if it's 0...but now I wonder why you want a 0 width and height div haha

revolv/templates/administrator/dashboard.html
```diff
@@ -3,143 +3,151 @@
...
-        <div class='small active administrator-button button'>Projects</div>
-    </a>
-    <div class='small primary administrator-button button'>Repayment</div>
-    <div class='small primary administrator-button button'>Ambassadors</div>
-    <a href="{%url "administrator:email" %}">
-        <div class='small primary administrator-button button'>Emails</div>
-    </a>
-</div>
+<div class="row">
+      <div class='row'>
```
noahsark769: Why do you have rows inside rows?

revolv/templates/administrator/dashboard.html
```diff
@@ -3,143 +3,151 @@
...
+            {% if user.revolvuserprofile.is_administrator %}
+              <div class='row'>
```
noahsark769: This indentation is two spaces while all the others are four. Whyyyyy

revolv/templates/administrator/dashboard.html
```diff
@@ -3,143 +3,151 @@
...
+              </div>
+            </li>
+          </ul>
+         </div>
         </div>
+      </div>
     </div>
-</div>
-
+        <br></br>
```
noahsark769: No no no no no no br! Use a margin instead

revolv/project/models.py
sameerav: for some reason it doesn't work if i just pass in text (the form returns a valueError)

revolv/templates/administrator/dashboard.html
noahsark769: Sorry that was a little bit fanatic haha. br just shouldn't be used for spacing between elements, that's what margin is for.

revolv/templates/donor/dashboard.html
```diff
@@ -3,79 +3,129 @@
 
 {% block title %}Home{% endblock %}
 
+{% block head %}
```
noahsark769: Oh god, it was a terrible decision to split the dashboard up into three separate templates. I'm sorry, my dashboard pull request will fix that.

revolv/project/views.py
```diff
@@ -155,20 +157,64 @@ def get_context_data(self, **kwargs):
...
+    def form_valid(self, form):
+        text = form.cleaned_data['update_text']
+        project = self.get_object()
+        project.add_update(text)
+        return super(PostProjectUpdateView, self).form_valid(form)
+
+    def form_invalid(self, form):
+        print("oh no", form)
+
+class EditProjectUpdateView(UserDataMixin, UpdateView):
```
noahsark769: Okay, so PostProjectUpdateView and EditProjectUpdateView are the exact same except for one line. There must be some way to DRY up this part!

revolv/templates/project/edit_project_update.html
```diff
@@ -0,0 +1,58 @@
...
+                <div class="medium-12 columns">
+                    {% include "base/partials/tooltip_formfield.html" with field=form.update_text %}
+                </div>
+            </div>
+            <button class="large right" id="save_project" type="submit">Save</button>
+        </form>
+    </div>
+</div>
+
+<!-- <div class='row'>
```
noahsark769: This can be deleted

revolv/templates/project/post_funding_update.html
```diff
@@ -31,7 +31,7 @@
                     {% include "base/partials/tooltip_formfield.html" with field=form.solar_url %}
                 </div>
             </div>
-            <button class="large right" id="save_project" type="submit">Save</button>
+            <button class="large right" id="save_project" type="create">Save</button>
```
noahsark769: I don't think "create" is an actual type...?

noahsark769: Also it looks like these templates are exactly the same except for this line, so they should also be DRY'd up

revolv/templates/project/project.html
```diff
@@ -2,8 +2,13 @@
 {% load staticfiles %}
 
 {% block head %}
+    <link rel="stylesheet" type="text/css" href="{% static "dashboard.css" %}" />
     <link rel="stylesheet" type="text/css" href="{% static "auth.css" %}">
     <link rel="stylesheet" type="text/css" href="{% static "project.css" %}">
+
+    <script type="text/javascript" src="{% static "js/vendor/d3.min.js" %}"></script>
+    <script type="text/javascript" src="{% static "js/vendor/svg4everybody.min.js" %}"></script>
```
noahsark769: Don't think you actually need svg4everybody here, it's just for the svg `use=...` stuff we have/used to have on the homepage

revolv/templates/project/project.html
```diff
@@ -17,53 +22,118 @@
...
-            {% for donation in project.donation_levels %}
-                <p>Description: {{ donation.description }}, Amount: {{ donation.amount }}</p>
-            {% endfor %}
-        {% endif %}
-        {% if project.is_completed or project.is_active and project.post_funding_updates %}
-            <h5>Post Funding Updates</h5>
-            <p>{{ project.post_funding_updates}}</p>
-        {% endif %}
+
+<div class="row dashboard">
```
noahsark769: If these are all absolutely positioned, I don't think there's any point in making them rows

revolv/templates/project/project.html
```diff
@@ -17,53 +22,118 @@
...
+                <p class ="num-donors">{{ donor_count }} donors</p>
+            </div>
+        </div>
+        <svg class="svg-graphics-container">
+            <circle class="outside-circle" cx="50%" cy="50%" r="125" fill="white" />
+        </svg>
+        <svg class="internal-graphics-container"></svg>
+        <div class="text-container">
+            <p class="percentage-text">{{ project.percent_complete }}%</p>
+            <!-- <p class="percentage-text">70%</p> -->
```
noahsark769: Deleteable

revolv/templates/project/project.html
```diff
@@ -17,53 +22,118 @@
...
+            <p class="funded">FUNDED</p>
+        </div>
+    </div>
+</div>
+
+<div class="heading row">
+    {% if project.cover_photo %}
+    <div class = "large-12 header-img-div" style='background-image: url("{{project.cover_photo.url}}")'>
```
noahsark769: Should be indented

noahsark769: Also: no spaces around the = here

revolv/templates/project/project.html
```diff
@@ -17,53 +22,118 @@
...
+    <div class="gray-bar"></div>
+    <div class="row categories">
+        <div class="category-text"><p class="category-text-inside">CATEGORIES   </p></div>
+        {% for category in project.categories %}
+            <div class="category-holder">
+                <p class="category-label">{{ category }}</p>
+            </div>
+        {% endfor %}
+
+        <!-- These two are in here for testing and viewing purposes: projects don't have categories right now. -->
```
noahsark769: They do though, see `Project.categories` (line 447 of project/models.py)

revolv/templates/project/project.html
```diff
@@ -17,53 +22,118 @@
...
+        <div class="category-holder"><p class="category-label">ARTS</p></div>
+        <div class="category-holder"><p class="category-label">HEALTH</p></div>
+    </div>
+    {% if project.video_url %}
+        <iframe width="100%" height="500" type="text/html" class="project_video youtube-player" frameborder="0" allowfullscreen>
+        </iframe>
```
noahsark769: No need for the space in between iframe tags

revolv/templates/project/project.html
```diff
@@ -17,53 +22,118 @@
...
+    {% endif %}
+
+    <div class="about-text"><p>{{ project.org_about }}</p>
     </div>
 </div>
 
+{% if updates|length > 0 %}
+<div class="row project-updates">
```
noahsark769: Indent

revolv/templates/project/project.html
```diff
@@ -17,53 +22,118 @@
...
+    {% if project.video_url %}
+        <iframe width="100%" height="500" type="text/html" class="project_video youtube-player" frameborder="0" allowfullscreen>
+        </iframe>
+    {% endif %}
+
+    <div class="about-text"><p>{{ project.org_about }}</p>
     </div>
 </div>
 
+{% if updates|length > 0 %}
```
noahsark769: Didn't we pass in `update_count` specifically for this reason?

revolv/templates/project/project.html
```diff
@@ -17,53 +22,118 @@
...
+            <div class="title-container">
+                <p>{{ project.title }}</p>
+            </div>
+            <div class="address-container">
+                <p>{{ project.location_street }}</p>
+                <p>{{ project.location_city_state_zip }}</p>
+            </div>
+        </div>
+
+        <!-- this donate button doesnt work because the model has a top attribute of like 4000 pixels -->
```
noahsark769: Can you add a TODO to this comment so we can be sure to find it later?

revolv/templates/project/project.html
```diff
@@ -17,53 +22,118 @@
...
+        <a href="#donate" data-reveal-id="donate-modal"><div class="donate-button-header">
+            <p>DONATE</p>
+        </div></a>
+    </div>
+</div>
+
+<div class="storage-div" id="{{ project.partial_completeness }}">
```
noahsark769: Should just be `storage`, `storage-div` is redundant

noahsark769: Also...what is this for? haha

revolv/project/views.py
sameerav: automatically validated by django as a mandatory field :)

I don't think any error is possible - any text string is a valid update right?

revolv/project/views.py
sameerav: wait aren't they both necessary? one has the template for creating a new update and the other has the template for editing an existing update? they both have the same form though

revolv/project/views.py
sameerav: it might just be something silly i'm doing, but the thing crashes if I delete those lines

noahsark769:
Hey @aykamko, can I get some help reviewing this pull request?

I made a ridiculous amount of comments haha. The project page looks really good, but this pull still needs a little bit of work. Specifically:

1. I didn't realize you were changing the dashboard page so much! It's not really needed given that we're going to update it to the other mockup this week. Also, it introduces a bunch of problems, including the fact that when you select an accordion item, the page scrolls down to it, which should not happen. Also, the text in the boxes is misaligned, and there is no link to the review page anymore :( Do you think it would be better to try to fix these issues or to revert the dashboard changes completely?
2. The project review page (/project/1/review) is somehow totally broken, you can't click on the buttons at the top. Not sure why this is, may be related to the modal again?
3. What's the plan for the project update pages (i.e. /project/1/update and /project/editupdate/1)? We're planning to put those into the dashboard right? If not, we should actually style the pages correctly.

Some things about the project page:
1. Can we make the overlay on the map darker? The text can be hard to read if there are a lot of streets on the map behind it.
2. I think we need to disable zoom on the map: as you're scrolling down the page, the map intercepts the scroll event and just zooms the map instead of scrolling the page past it. One way to do this is to put a transparent overlay over the map, since you can still scroll on the dark overlay right now.
3. Looks like project update styles are a bit weird on tablet: 
screen shot 2015-04-02 at 8 50 10 pm
4. The spedometer has higher z-index than the nav bar, which should not happen: 
screen shot 2015-04-02 at 8 51 38 pm
5. We need to add in actually rendering the real associated categories, since we have them now
6. We should redraw the spedometer svg circle stuff on window resize: if you transition from landscape to portrait on mobile then it gets off
7. Portrait mobile styles definitely need some work: 
screen shot 2015-04-02 at 8 54 30 pm
screen shot 2015-04-02 at 8 54 36 pm
screen shot 2015-04-02 at 8 54 49 pm
8. On landscape mobile, we need to resize the iframe since it's getting cut off: 
screen shot 2015-04-02 at 8 55 11 pm

I think that should be everything. Definitely let me know if you have questions about anything I said, there's a LOT haha

revolv/project/models.py
noahsark769: But like, why can't you call the field `text` instead of `update_text`? `update_text is pretty redundant`

revolv/project/tests/test_models.py
noahsark769: Why is this test deprecated? We don't actually have a login modal and we still want it to redirect to the signin page if they try to donate anonymously right?

revolv/project/views.py
noahsark769: Hm I guess you're right. We should probably just raise a 500 error in this case, to be honest.

revolv/project/views.py
noahsark769: No, I mean delete the comment haha. Both views are still necessary

noahsark769: I think we don't really need a comment here because it's pretty self explanatory why the attributes of this view are what they are. However, you can leave the comment in if you expand on why it's not self explanatory why you might need the template and form there.

revolv/project/views.py
noahsark769: Weird...it might be the fact that you're mixing in UserDataMixin BEFORE UpdateView, and super isn't getting called correctly in one? Not sure

revolv/static/js/base/project.js
sameerav: wait why can this be deleted?
doesn't it draw the circle after the document is ready?

revolv/static/mixins.scss
sameerav: That makes sense actually. A lot of the same code is copied in modal.scss already - do you just want me to delete duplicate code from mixins.scss?

revolv/static/project.scss
sameerav: yeah! Foundation by default adds a margin to the rows so I wanted to make them fullscreen.

revolv/static/project.scss
sameerav: To override the foundation settings yeah :(

revolv/static/project.scss
sameerav: It's always a div with the image set as the background image!

revolv/static/project.scss
sameerav: they're all different for the most part though. Do you think it would it be worth creating variable names if each one has its own variable anyway?

revolv/static/project.scss
sameerav: the svg-graphics-container has the partial circle in it and the internal-graphics-container has the text with the funding percentage

revolv/static/project.scss
sameerav: Moved color to parent! They do need to be position: absolute because otherwise they push each other around :(

revolv/static/project.scss
sameerav: LOL tbh not sure why I did that - I'll reposition it in a way that makes more sense

revolv/templates/administrator/dashboard.html
sameerav: I just reverted all the dashboard pages to what's on master right now!

revolv/templates/project/project.html
sameerav: The top one isn't absolute and one of the ones in the middle isn't absolute either

revolv/templates/project/project.html
sameerav: oh what i meant was that this particular project didn't have categories attached to it specifically!

revolv/templates/project/project.html
sameerav: nah that was donor_count!

revolv/templates/project/project.html
sameerav: I'm sure there's a non-hacky way to do this but I just did this so i could get the partial completeness in the javascript

how would i access it without the hackyness of this div?

revolv/static/mixins.scss
noahsark769: Yup, anything that you can clean up is worth cleaning up

revolv/static/project.scss
noahsark769: Bleh

revolv/templates/project/project.html
noahsark769: Oh alright, i guess just the absolute ones dont need to be rows then

revolv/static/js/base/project.js
noahsark769: No I mean the newline haha

revolv/static/project.scss
noahsark769: Oh okay I guess that's alright, it will still cause problems if you have rows within rows though I think?

revolv/static/project.scss
noahsark769: I was just thinking that it might be nice to have all the offsets in one place so that someone reading the code can see what's going on

revolv/static/project.scss
noahsark769: Ah

revolv/templates/project/project.html
noahsark769: Oh yeah lol

revolv/project/forms.py
```diff
@@ -65,16 +65,15 @@ class Meta:
...
 
-
-class PostFundingUpdateForm(forms.ModelForm):
-    """
-    A form for providing post funding updates about a project
+class EditProjectUpdateForm(forms.ModelForm):
+    """ 
+    A form used to edit updates about a project
     """
+    #SET SOME KIND OF DEFAULT VALUE
```
noahsark769: I think we can still get rid of this line

revolv/project/models.py
```diff
@@ -336,6 +330,32 @@ def amount_donated_organically(self):
         )["amount__sum"] or 0.0
 
     @property
+    def location_street(self):
+        """
+        :return: a string of the street name of the location of this project
```
noahsark769: I think we should mention in these docstrings that the function will return "" if the location is malformed

revolv/project/models.py
```diff
@@ -426,6 +455,34 @@ def categories(self):
         return [category.title for category in self.category_set.all()]
 
     @property
+    def updates(self):
+        """
+        :return: The set of all ProjectUpdate models associated with this project.
+        """
+        #return self.update_set.all()
```
noahsark769: We can delete this line instead of commenting it out

revolv/project/models.py
```diff
@@ -426,6 +455,34 @@ def categories(self):
...
+    
+    date = models.DateField(
+        'Date of update creation',
+        help_text = "What time was your update created?",
+        auto_now_add = True
+    )
+    
+    project = models.ForeignKey(
+        Project,
+        related_name="update"
```
noahsark769: I think "updates" might be a better related name, `project.update.all()` is a little weird

revolv/project/tests/test_integration.py
```diff
@@ -0,0 +1,193 @@
...
+        self.assertUserAuthed(response)
+        project = Project.factories.base.create()
+        response = self.client.post('/project/%d/update' % project.pk, {'update_text': 'This is an update'})
+        
+        # the deny_access method is called in the view from the userdatamixin, which returns a redirect to the homepage
+        self.assertEqual(response.status_code, 302)
+        
+class EditProjectUpdatesTest(TestUserMixin, UserTestingMixin, TestCase):
+
+    def assert_make_update(self, project, text):
```
noahsark769: Sorry to be pedantic about this, but the point of this function isn't really to assert anything. Sure, it asserts that the user is authed at one point, but the point is to make a new update, so I think it's fine to call it `make_update`. I think that methods whose point is to check that something is the same as expected (or even check that more than one thing is the same as expected) should start with `assert_` but methods like this don't need to.

revolv/project/tests/test_integration.py
```diff
@@ -0,0 +1,193 @@
...
+        Returns the update that was just created.
+        """
+        response = self.send_test_user_login_request()
+        self.assertUserAuthed(response)
+        #response = self.client.post('/project/%d/update' % project.pk, {'update_text': text})
+        ProjectUpdate(project = project, update_text = text).save()
+        new_updates = ProjectUpdate.objects.filter(update_text = text)
+        return new_updates[0]
+
+    def assert_test_edit_update(self, donor = False):
```
noahsark769: This one definitely SHOULD start with `assert_` but having `test` in the name also makes it confusing. How about `assert_user_can_or_cant_make_updates` or `assert_user_update_permissions_work`?

revolv/project/tests/test_models.py
```diff
@@ -11,6 +12,31 @@
...
+        project = Project.factories.base.create()
+        update1 = ProjectUpdate(update_text = 'This is update text', project=project)
+        update2 = ProjectUpdate(update_text = 'This is another update', project=project)
+
+        # tests basic construction
+        self.assertEqual('This is update text', update1.update_text)
+        self.assertEqual('This is another update', update2.update_text)
+
+        # tests project relationship
+        self.assertEqual(update1.project_id, update2.project_id)
```
noahsark769: Again, shouldn't you be comparing `update1.project` rather than `update1.project_id`? Is `update1.project_id` even a thing?

revolv/project/views.py
```diff
@@ -155,19 +157,42 @@ def get_context_data(self, **kwargs):
...
+        return super(PostProjectUpdateView, self).form_valid(form)
+
+
+class EditProjectUpdateView(TemplateProjectUpdateView):    
+    model = ProjectUpdate
+
+    def get_success_url(self):
+        return reverse('project:review', kwargs={'pk': self.get_object().project_id})
+
+    def form_valid(self, form):
```
noahsark769: Man these functions look so nice now, +1

revolv/static/dashboard.scss
```diff
@@ -61,14 +55,113 @@
     }
 }
 
-.proj-button {
-    width: 100%;
-    float: none !important;
+.project-container {
+    padding: 0px;
```
noahsark769: If 0, just use `0`, not `0px`. This is general rule of thumb for CSS style :)

revolv/static/home.scss
```diff
@@ -277,7 +277,7 @@ body {
         left: 0;
         right: 0;
         bottom: 0;
-        height: 50%
+        height: 50%;
```
noahsark769: Nice!

revolv/static/js/base/dashboard.js
```diff
@@ -0,0 +1,43 @@
+$(document).ready(function() {
```
noahsark769: I thought we got rid of the dashboard changes, so we don't need this file, right?

revolv/static/js/base/partialcircle.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * Draws a partial circle.
```
noahsark769: Thanks for writing this docstring, I owe you one :D

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
+$(document).ready(function () {
+	draw()
```
noahsark769: Did jshint not catch this? You're missing a semicolon and this line is indented with a tab instead of a space

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+    $(".outside-circle" ).remove();
+    draw();
+});
+
+var setTextSize = function(radius) {
+    
+    var percentage_size = 3*radius/100;
+    var funded_size = percentage_size*0.5;
+
+    // d3.select(".percentage-text").attr("font-size", percentage_size + "rem")
```
noahsark769: All these commented out lines we can just delete

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+
+    // if ($(document).width() > 200 && $(document).width() < 800) {
+    //     radius = 48;
+    // }
+    var radius = $(document).width()*0.12;
+    setTextSize(radius)
+
+    resizeIframe();
+
+    // <circle class="outside-circle" cx="50%" cy="50%" r="" fill="white" />
```
noahsark769: This comment is okay for a reference I think, but the other comments in this function can be deleted

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+
+var draw = function() {
+    // var radius = 100,
+    //     padding = 10;
+
+
+    // if ($(document).width() > 200 && $(document).width() < 800) {
+    //     radius = 48;
+    // }
+    var radius = $(document).width()*0.12;
```
noahsark769: Need spaces before and after the * here

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+
+    // percentage_text.css("font-size", percentage_size + "rem")
+    // funded_size.css("font-size", funded_size + "rem")
+}
+
+var resizeIframe = function() {
+    d3.select("iframe").attr("height", $(document).width()*0.4 + "")
+}
+
+var draw = function() {
```
noahsark769: This function should have a docblock too :)

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+
+    // <circle class="outside-circle" cx="50%" cy="50%" r="" fill="white" />
+
+    var outsideCircle = d3.select(".svg-graphics-container")
+        .append("circle")
+        .attr("class", "outside-circle")
+        .attr("cx", "50%")
+        .attr("cy", "50%")
+        .attr("r", radius + $(document).width() * 0.03 + "")
+        .attr("fill", "white")
```
noahsark769: Missing semicolon, does jshint even work? lol

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+
+    if (radius > 100){
```
noahsark769: Space after )

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+        radius = 100;
+        outsideCircle.attr("r", 125)
```
noahsark769: Missing semicolon

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    if (radius < 60) {
+        radius = 60;
+        outsideCircle.attr("r", 75)
```
noahsark769: Missing semicolon

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    setTextSize(radius)
```
noahsark769: ;

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+
+
+    var padding = radius*0.1;
```
noahsark769: Space around *

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+
+
+    // this line will get actual partial completeness
+    var partialCompleteness = window.PARTIAL_COMPLETENESS;
+
+    // use this for test purposes so you can see the progress bar
+    var partialCompleteness = 0.7;
```
noahsark769: Don't you need to comment this out? -_-

noahsark769: You can just delete it, to be honest haha. If it's there, people will just forget to comment it out when they're done testing

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+    var svg = d3.select(".internal-graphics-container")
+        .attr("width", dimension)
+        .attr("height", dimension)
+        .append("g");
+
+    var stroke = radius*0.2
+
+    var circleGrouping = svg.append("g").attr("class", "project-badge-circle-grouping").attr("stroke-width", stroke + "px");
+    var partialGrouping = svg.append("g").attr("class", "project-badge-partial-grouping").attr("stroke-width", stroke + "px");
+
```
noahsark769: There are lots of places in this file where there are two newlines when there only need to be one. Can we try to consolidate them?

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+    var dimension = (2 * radius) + (2 * padding);
+
+    var translateVar = (radius + padding)*0.5;
+
+    var svg = d3.select(".internal-graphics-container")
+        .attr("width", dimension)
+        .attr("height", dimension)
+        .append("g");
+
+    var stroke = radius*0.2
```
noahsark769: Spaces around *

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+    // d3.select(".percentage-text").attr("font-size", percentage_size + "rem")
+    // d3.select(".funded").attr("font-size", funded_size + "rem")
+
+    d3.select(".percentage-text").attr("style", "font-size:" + percentage_size + "rem")
+    d3.select(".funded").attr("font-size", "font-size:" + funded_size + "rem")
+
+
+    // percentage_text.css("font-size", percentage_size + "rem")
+    // funded_size.css("font-size", funded_size + "rem")
+}
```
noahsark769: ;

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+
+var resizeIframe = function() {
+    d3.select("iframe").attr("height", $(document).width()*0.4 + "")
+}
```
noahsark769: ;

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+var draw = function() {
+    // var radius = 100,
+    //     padding = 10;
+
+
+    // if ($(document).width() > 200 && $(document).width() < 800) {
+    //     radius = 48;
+    // }
+    var radius = $(document).width()*0.12;
+    setTextSize(radius)
```
noahsark769: ;

revolv/static/js/base/project.js
```diff
@@ -0,0 +1,97 @@
...
+
+    // this line will get actual partial completeness
+    var partialCompleteness = window.PARTIAL_COMPLETENESS;
+
+    // use this for test purposes so you can see the progress bar
+    var partialCompleteness = 0.7;
+
+    var dimension = (2 * radius) + (2 * padding);
+
+    var translateVar = (radius + padding)*0.5;
```
noahsark769: Spaces around * lol

revolv/static/mixins.scss
```diff
@@ -1,3 +1,6 @@
+@import "foundation-settings";
```
noahsark769: Why do we need to import these here?

revolv/static/project.scss
```diff
@@ -43,6 +43,405 @@
...
+    }
+
+    &.about, &.project-updates, &.dashboard {
+        margin-right: 15%;
+        margin-left: 15%;
+        max-width: 70%;
+    }
+
+    &.dashboard {
+        // height: 600px;
```
noahsark769: Just delete instead of commenting out :)

revolv/static/project.scss
```diff
@@ -43,6 +43,405 @@
...
+            right: 5%;
+            transition: all 0.3s;
+
+            .donate-link {
+                width: 100%;
+                height: 100%;
+                display: inline-block;
+
+                p {
+                    //margin: 5% 0px;
```
noahsark769: Just delete instead of commenting out

revolv/static/project.scss
```diff
@@ -43,6 +43,405 @@
...
+        .svg-graphics-container {
+            // this contains the formatting for the white outside circle - this works
+            position: absolute;
+            height: 100%;
+            width: 100%;
+        }
+
+        .internal-graphics-container {
+            position: absolute;
+            // height: 100%;
```
noahsark769: Just delete instead of commenting out :)

revolv/static/project.scss
```diff
@@ -43,6 +43,405 @@
...
+    }
+}
+
+.about {
+    margin-top: 200px;
+
+    .header-text {
+        p {
+            font-weight: 800;
+            color: #444444;
```
noahsark769: These can be shortened to three digits e.g. #444, #999 etc

revolv/static/project.scss
```diff
@@ -43,6 +43,405 @@
...
+            height: 50px;
+            width: 100%;
+            text-align: center;
+            color: $revolv-dark-blue;
+            font-weight: 300;
+            line-height: 0.5;
+
+            .date-year {
+                font-size: 4.5rem;
+                margin-bottom: 0rem;
```
noahsark769: Just 0, no units

revolv/static/project.scss
```diff
@@ -43,6 +43,405 @@
...
+        }
+    }
+
+    .donate-button-header {
+        border: 2px solid white;
+        width: 80%;
+        height: 70px;
+        z-index: 100;
+        margin: 305px 10% 25px 10%;
+        transition: all 0.3s;
```
noahsark769: Do you need prefixes -webkit-transition, -moz-transition etc here?

revolv/static/project.scss
```diff
@@ -77,3 +476,227 @@
     height: 300px;
     margin: 0 20px 20px 20px;
 }
+
+// --- Mobile Responsiveness Styles ------------------------------------------------------
+// TO FIX @MEDIA HEADERS AND TO ADD IN PORTRAIT/ORIENTATION STUFF
```
noahsark769: You mean TODO?

revolv/static/project.scss
```diff
@@ -77,3 +476,227 @@
...
+
+            .update-text {
+                font-weight: 500;
+                font-size: 1rem;
+            }
+        }
+    }
+    .map {
+
+        height: 250px;
```
noahsark769: This might make a good variable e.g. `$map-height`, you're using it in a few different places

revolv/static/project.scss
```diff
@@ -77,3 +476,227 @@
...
+                font-size: 1rem;
+                font-weight: 800;
+            }
+        }
+    }
+
+    .dashboard {
+        height: 700px;
+        z-index: 1;
+        top: 0px;
```
noahsark769: Just 0, no units

revolv/static/project.scss
```diff
@@ -77,3 +476,227 @@
...
+
+        .overlay {
+            
+            .donate-button-header {
+                left: 15%;
+                width: 70%;
+                height: 15%;
+                top: 170%;
+                border-color: $revolv-dark-blue;
+                // width: 0;
```
noahsark769: Just delete, don't comment out

revolv/static/project.scss
```diff
@@ -77,3 +476,227 @@
...
+                        line-height: 40px;
+                        color: $revolv-dark-blue;
+                    }
+                }
+            }
+
+            .project-title {
+                font-size: 1.5rem;
+                line-height: 1.5rem;
+                // top: 20px;
```
noahsark769: Just delete, don't comment out

revolv/static/project.scss
```diff
@@ -77,3 +476,227 @@
...
+        .about-text {
+            p {
+                font-size: 1rem;
+            }
+        }
+
+    }
+
+    .map {
+        .title-container {
```
noahsark769: Didn't you already set these styles above?

revolv/static/project.scss
```diff
@@ -77,3 +476,227 @@
...
+        .donate-button-header {
+            p {
+                font-size: 1.5rem;
+            }
+        }
+    }
+}
+
+@media #{$medium-only} {
+    .dashboard {
```
noahsark769: This is a huge chain of hierarchies. Can't you just do `.statistic { ... }` instead?

revolv/templates/donor/dashboard.html
```diff
@@ -21,7 +21,7 @@
     {% endif %}
 </div>
 <div class='row'>
-	<div class='small-3 columns dash-box'>
+  <div class='small-3 columns dash-box'>
```
noahsark769: Why indent with two spaces? everything else in indented with 4

noahsark769:
Dude, WOW! The responsive styles look AWESOME!!

Still a few things to do, though mostly cosmetic at this point. I made a lot of comments haha. Is your jshint running every time you commit? If not, you should do `pre-commit install`.

1. The map overlay styles still need a tiny bit of tuning up for mobile portrait and landscape (especially portrait): the address needs a little more line height (lines are too close together) and the DONATE is not vertically aligned, I don't think:
screen shot 2015-04-09 at 8 44 47 pm
screen shot 2015-04-09 at 8 44 57 pm
2. One small feature: do you think we could preserve line breaks in the project update text (i.e., break every time we see a newline character into separate <p> tags)? Right now, even if the person enters in newlines when writing the update, it just shows up as a large block of text, presumably because HTML doesn't distinguish between 1 character and 2 characters of whitespace and just collapses everything down to one space.

revolv/project/tests/test_models.py
sameerav: yeah update1.project_id returns the same result as update1.project_pk or update1.project.pk

i think it exists so you could give things custom ids??? so not stuff relevant to this

but will change

revolv/static/js/base/dashboard.js
sameerav: oops yeah

revolv/static/project.scss
sameerav: i added in -moz-transform and -webkit-transform because safari needed them for some reason

safari doesn't seem like it needs -moz-transition or -webkit-transition but i'll add them in to be safe

revolv/static/project.scss
sameerav: whoops yeah deleted now though

revolv/static/project.scss
sameerav: nah the font size got smaller

revolv/static/project.scss
sameerav: for some weird reason the styles don't apply without the long chain of hierarchies - do you know why?

noahsark769: It's probably because of how [CSS Specificity](http://www.smashingmagazine.com/2007/07/27/css-specificity-things-you-should-know/) works. Should be fine for now.

revolv/static/project.scss
noahsark769: Hm alright, actually according to [this](http://caniuse.com/#feat=css-transitions) transition is supported in all major browsers right now, except for IE 9 and 8 in which -webkit and -moz won't help. So I guess this is fine.

revolv/static/project.scss
noahsark769: Oh kk

revolv/static/mixins.scss
```diff
@@ -1,3 +1,5 @@
+@import "bower_components/foundation/scss/foundation";
```
noahsark769: I am 99% sure you don't need this right?

revolv/templates/donor/dashboard.html
```diff
@@ -74,8 +74,8 @@ <h4 class='subheader text-center'>My Projects</h4>
             {% empty %}
                 <p> You have not donated to any projects. </p>
             {% endfor %}
-		</div>
-	</div>
+    </div>
+  </div>
```
noahsark769: Two spaces still?

noahsark769:
PREPARE TO GO TO SPACE

Issue #211, opened 2015-03-31
Random bugs with the homepage on mobile


noahsark769:
1. Margin at the top is too large
2. Times new roman displays in the svg

andylouisqin:
@noahsark769 this too should be fixed by #218 heh

noahsark769:
```
            _
           /(|
          (  :
         __\  \  _____
       (____)  `|
      (____)|   |
       (____).__|
        (___)__.|_____
```

PR #214, opened 2015-04-05
FIxes issue #211 (mobile margin too large, svg's with times font)

andylouisqin:


noahsark769:
Alright let's see if this works

PR #215, opened 2015-04-05
Monthly donation email hook

### LINKED ISSUE #125: Monthly engagement email hook
See description in #51. No html email, just text based and the job to send these out. Might make sense to brainstorm some kind of admin interface for doing this too

erictshen:
Fixes #125 

Added a command for running monthly donation email jobs. Will add a job to heroku when this gets merged in.

@andylouisqin will also be working on email text once this gets merged in.

Example email from the terminal.
screen shot 2015-04-05 at 1 59 38 pm


revolv/base/management/commands/monthlydonationemail.py
```diff
@@ -0,0 +1,49 @@
+import datetime
+
+from django.core.management.base import BaseCommand
+
+from revolv.base.models import RevolvUserProfile
+from revolv.lib.mailer import send_revolv_email
+
+
+class Command(BaseCommand):
```
noahsark769: Can we call this something semantic? How about `MonthlyDonationEmailCommand`?

revolv/base/management/commands/monthlydonationemail.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    This handle function is run when the command "python manage.py monthlydonationemail"
+    is typed into the command line.
+
+    It iterates through all users, and sends a monthly donation email to any user
+    who has donated in the past month
```
noahsark769: Nitpick: should be a period after this line

revolv/base/management/commands/monthlydonationemail.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+
+    def handle(self, *args, **options):
+        if datetime.date.today().day == 1:  # checks if the it is the first day of the month
```
noahsark769: Can we have something be printed in the other case?

revolv/base/management/commands/monthlydonationemail.py
erictshen: Django expects something called Command =(

revolv/base/management/commands/monthlydonationemail.py
```diff
@@ -0,0 +1,49 @@
...
+                    user = revolv_user_profile.user
+                    context['user'] = user
+                    context['payments'] = donation_set
+                    send_revolv_email(
+                        'monthly_donation_email',
+                        context, [user.email]
+                    )
+
+
+"""
```
noahsark769: This docstring should be on the line after the function definition

revolv/templates/emails/emails.yml
```diff
@@ -42,3 +42,35 @@ post_donation:
...
+    body:
+        "
+        Hi {{user.first_name}}!
+
+        Thank you for your invaluable donations to Re-volv! Each contribution is a major step in bringing clean energy to everyone.
+
+        Monthly Stats
+
+        {% for payment in payments %}
+        You donated {{payment.amount}} to {{payment.project.title}}
```
noahsark769: Should at least have a period

noahsark769:
We need tests!

erictshen:
Tests currently fail. Main reason why is that we can't override the created_at variable for the payments.

https://github.com/rbarrois/factory_boy/issues/102

Will most likely implement this the first way mentioned in that link, unless you'd prefer some other way of fixing this issue.

erictshen:
Updated factory boy so we can override the created_at attribute! Passes tests now, let me know if you would like more test coverage.

revolv/base/management/commands/monthlydonationemail.py
noahsark769: Oh bleh, okay

revolv/lib/tests/test_mailer.py
```diff
@@ -62,3 +68,77 @@ def test_cc_admins(self):
...
+        """
+        # creates 2 users and 2 projects
+        user1, user2 = RevolvUserProfile.factories.base.create_batch(2)
+        project1, project2 = Project.factories.base.create_batch(2)
+
+        # gets a day from the last month
+        date_of_last_month = datetime.date.today().replace(day=1) - datetime.timedelta(days=1)
+
+        # creates two payments and sets their created_at date to a date from last month
+        Payment.factories.base.create(
```
noahsark769: You actually don't need to specify all of these values, you can just say `Payment.factories.donation.create(created_at=...)`. Everything else doesn't really matter, so no need to include it

noahsark769:
@erictshen I think we should add two more tests: (1) one where we mock out the return value of `.now()` and make sure that we can run the command with override=False, and (2) one where we run the command from some random day of the month that's not the first day, and make sure that only the donations from the previous MONTH get tracked (not donations from the past 30 DAYS, which may be different than the past month).

Also, it might make sense to send the admins an email when this job is done running, like "An automatic batch of monthly donation engagement emails was sent" or something like that. What do you think - should be pretty easy to do, since we don't need a fancy email template and I believe we have a utility function to get all administrator emails already.

revolv/base/management/commands/monthlydonationemail.py
```diff
@@ -0,0 +1,61 @@
...
+                    context = {}
+                    user = revolv_user_profile.user
+                    context['user'] = user
+                    context['payments'] = donation_set
+                    send_revolv_email(
+                        'monthly_donation_email',
+                        context, [user.email]
+                    )
+        else:
+            print "Attempted to run monthlydonationemail command on a day that was not the 1st of the month"
```
noahsark769: Can we say that the command was not actually run? Or something about using --override to force it? Also we should at least end this sentence with a period :)

erictshen:
Added additional test coverage. Also added a command line arguments to email administrators when the monthlydonationemail command is run, I can make it default behavior if you'd prefer.

noahsark769:
Two more things!

1. I think that emailing admins should be the default behavior
2. I think that in the docstring for the Command, we should specify the functionality about only emailing people who have donated within the previous month, not the previous 30 days.

New tests look good!

noahsark769:
1. Need to add tests to check that emails to admins are sent
2. Admin email template should include the number of donation emails sent in the batch, I think

noahsark769:
Ship it!
```
 --    .-""-.
   ) (     )
  (   )   (
     /     )
    (_    _)                     0_,-.__
      (_  )_                     |_.-._/
       (    )                    |_--..\
        (__)                     |__--_/
     |''   ``\                   |
     |        \                  |      /b.
     |         \  ,,,---===?A`\  |  ,==y'
   ___,,,,,---==""\        |M] \ | ;|\ |>
           _   _   \   ___,|H,,---==""""bno,
    o  O  (_) (_)   \ /          _     AWAW/
                     /         _(+)_  dMM/
      \@_,,,,,,---=="   \      \\|//  MW/
--''''"                         ===  d/
                                    //
                                    ,'__________________________
   \    \    \     \               ,/~~~~~~~~~~~~~~~~~~~~~~~~~~~
                         _____    ,'  ~~~   .-""-.~~~~~~  .-""-.
      .-""-.           ///==---   /`-._ ..-'      -.__..-'
            `-.__..-' =====\\\\\\ V/  .---\.
 PGMG                 ~~~~~~~~~~~~, _',--/_.\  .-""-.
                            .-""-.___` --  \|         -.__..-
```

Issue #165, opened 2015-03-01
Validate payments using the paypal API AND actually charge cards on prod


noahsark769:
We need to actually do this in the payments backend, and set up paypal to actually work on production. Right now, we don't validate and just throw an exception (500 error) if the card data is not valid. Let's make sure we use sandbox paypal api keys on staging and local also

noahsark769:
Fixed by #216

PR #216, opened 2015-04-05
paypal charge

aykamko:
#165


revolv/payments/lib/instruments.py
```diff
@@ -59,12 +66,14 @@ def __init__(self, credit_card):
 
     def charge(self, amount):
         """
-        Charge the credit card for the given amount.
+        Documentation: https://github.com/paypal/PayPal-Python-SDK
+
+        Charge the credit card for the given amount, floored to 2 decimal
+        places. Raises InvalidCreditCardException if charge fails.
 
         :param amount:
```
noahsark769: Say that this needs to be a float :)

noahsark769: Or at least add a test that makes sure it works with both ints and floats

aykamko: Note to self: add tests for int, float, and Decimal

aykamko:
Note to self: add gmail and Paypal Dev credentials to documentation before closing pull

revolv/project/tests/tests_general.py
```diff
@@ -298,6 +298,7 @@ def perform_valid_donation(self):
             HTTP_X_REQUESTED_WITH='XMLHttpRequest'
         )
 
+    @override_settings(ENABLE_PAYMENT_CHARGING=False)
```
noahsark769: We should have a comment saying why this is necessary to override (i.e. that dummy card numbers will cause the paypal sdk stuff to fail)

revolv/templates/project/partials/payment_modal.js
```diff
@@ -378,6 +389,13 @@ $('#donate-form').submit(function(e) {
         $donateModalErrors.children('.error-msg')
             .text('Your card information is invalid. Please correct it.');
         $('#donate-modal').foundation('reveal', 'open');
+    }).always(function () {
+        paymentSpinner.stop();
+        setTimeout(function() {
```
noahsark769: This deserves a comment saying why we need to do setTimeout (i.e. that if we don't the modal fading animation overlaps)

revolv/settings.py
```diff
@@ -295,4 +291,5 @@
     ('en-us', 'English'),
 ]
 
-CHARGE_INSTRUMENT = False
+# be careful!
```
noahsark769: I think we should expand this comment to talk about how the paypal keys work and how cards won't actually get charged unless there are non-sandbox keys. Also should put that somewhere more visible like the readme AND the docstring of PaymentInstrument

noahsark769:
@aykamko seems like this PR should be good to go now, can you merge in master though?

noahsark769:
NOICE

PR #217, opened 2015-04-07
Quick bug fix for category select

### LINKED ISSUE #196: Category select box has no error display
If you try to submit a project edit with no categories, it simply goes back to the same page and doesn't tell you anything. I realize this is because chosen has styles everywhere, but we should figure out some way to fix it

erictshen:
Fixes #196 

Two changes. Made it so that required=False for the field corresponding to category select. Also had it so the html would display any errors.

(Example screenshot)
screen shot 2015-04-06 at 7 02 46 pm


noahsark769:
Aw yeah revolv++

PR #223, opened 2015-04-14
New dashboard styles

noahsark769:
IT'S THE MOMENT WE'VE ALL BEEN WAITING FOR!

Things that still don't work:
1. Only the description tab is implemented, we need posting updates, funding stats, and statistics
2. Need to write a suite of WebTests for various dashboard flows: (a) making a donation for the first time and a second time, (b) proposing a project, (c) approving a project
3. Change password page on the dash
4. Better looking page for donors who have not donated to any projects yet.
5. Use `Project.org_about_lines` on the project page
6. We need to use the url param `active_project` (i.e. `http://localhost:8000/dashboard/admin/?active_project=3`) correctly, and write a test to make sure that works
7. Make the post updates button go to the right place
8. The nav bar menu doesn't appear on the dash page for some reason and I'm not sure why
9. Need to make everything work on mobile
10. Need a way to undo an "approve" action

screen shot 2015-04-14 at 5 12 03 am
screen shot 2015-04-14 at 5 18 39 am
screen shot 2015-04-14 at 5 18 50 am
screen shot 2015-04-14 at 5 19 06 am
screen shot 2015-04-14 at 5 19 51 am

I had to make a few changes from the mockup (most notably adding a create project button, because there wasn't one) and I think some of the styles could still be a little bit better, but in interest of time I think this is great for now. It actually turned out pretty nice, I think! As far as I can tell, all the functionality we had on the old dashboard should work still.

@andylouisqin @sameerav @erictshen I need all your eyes on this! Please review it before Tuesday night if you can! @vivekraghuram your input would be helpful as well! It's currently 5:17am and I would like to go to sleep, but I'll definitely be available tomorrow to update this. [Throwback to my last Blueprint late night PR](https://github.com/calblueprint/alt-breaks/pull/7).

Thanks guys!


revolv/templates/base/partials/dashboard_sidebar.html
```diff
@@ -0,0 +1,75 @@
...
+<div class="role-select-options">
+    <div class="dashboard-header">View as:</div>
+    {% if user.revolvuserprofile.is_administrator %}
+        <div class="role-select-option"><a href="/dashboard/admin/">administrator</a></div>
+        <div class="role-select-option"><a href="/dashboard/ambassador/">ambassador</a></div>
+        <div class="role-select-option"><a href="/dashboard/donor/">donor</a></div>
+    {% elif user.revolvuserprofile.is_ambassador %}
+        <div class="role-select-option"><a href="/dashboard/ambassador/">ambassador</a></div>
+        <div class="role-select-option"><a href="/dashboard/donor/">donor</a></div>
+    {% endif %}
```
erictshen: Could include ambassador and donor by default, and only have an if for checking administrator.

revolv/templates/base/partials/dashboard_sidebar.html
```diff
@@ -0,0 +1,75 @@
...
+        <div class="dashboard-header">No projects found.</div>
+    {% else %}
+        {% for project_group_name, project_list in project_dict.items %}
+            <div class="dashboard-header">{{ project_group_name.0 }}</div>
+            {% with outer_counter=forloop.counter %}
+                {% for project in project_list %}
+                    <div class="dashboard-sidebar-project-container dashboard-data-link {% if outer_counter == 1 and forloop.counter == 1 %} active{% endif %}" data-section="project-{{project.pk}}">
+                        <div class="dashboard-sidebar-bullet-point {{project_group_name.1}}"></div>
+                        <div class="dashboard-sidebar-project {{project_group_name.1}} dashboard-data-link">{{project.title}}</div>
+                    </div>
```
erictshen: If we're doing the tuple route with project_group_name and html class, make sure to comment somewhere (preferably in both the view and this html) about what they are.

revolv/static/js/base/dashboard.js
```diff
@@ -0,0 +1,22 @@
...
+            $(".role-select-options").slideUp();
+            $this.data("state", "collapsed");
+            $this.removeClass("fa-rotate-180");
+        } else {
+            $(".role-select-options").slideDown();
+            $this.data("state", "expanded");
+            $this.addClass("fa-rotate-180");
+        }
+    });
+
```
erictshen: Add a comment for how you're switching between projects!

revolv/administrator/views.py
```diff
@@ -12,15 +11,17 @@
...
         context = super(AdministratorDashboardView, self).get_context_data(**kwargs)
-        context['proposed_projects'] = Project.objects.get_proposed()
-        context['active_projects'] = Project.objects.get_active()
-        context['completed_projects'] = Project.objects.get_completed()
-        context['all_projects'] = list(chain(context['proposed_projects'],
-                                             context['active_projects'], context['completed_projects']))
+        project_dict = {}
+        project_dict[('Proposed Projects', "proposed")] = Project.objects.get_proposed()
+        project_dict[('Active Projects', "active")] = Project.objects.get_active()
+        project_dict[('Completed Projects', "completed")] = Project.objects.get_completed()
```
erictshen: Using tuples is kind of hacky, could just have the display text as a key, and have some helper or dictionary to convert that to the corresponding css class.

It's fine to keep it this way, but add some comments!

revolv/templates/base/dashboard.html
```diff
@@ -0,0 +1,85 @@
...
+                {% if first_project %}
+                    <p>Click <a href="{% url "project:view" pk=first_project.pk %}">here</a> to start contributing.</p>
+                {% endif %}
+            </div>
+        {% endif %}
+        {% for project_group_name, project_list in project_dict.items %}
+            {% with outer_counter=forloop.counter %}
+                {% for project in project_list %}
+                    <!-- {{outer_counter}} -->
+                    <!-- {{ forloop.counter}} -->
```
erictshen: Extra comments to delete.

erictshen:
The logic and code look fine (some minor notes though). Make sure to document / comment this more sometime, esp. for some of the html logic / classes. But looks good from my end!

revolv/templates/base/partials/dashboard_sidebar.html
noahsark769: Oh you're right, good catch!

revolv/ambassador/views.py
```diff
@@ -2,21 +2,31 @@
...
+
+        project_dict = {}
+        project_dict[ProjectGroup('Drafted Projects', "drafted")] = Project.objects.get_drafted(user_projects)
+        project_dict[ProjectGroup('Proposed Projects', "proposed")] = Project.objects.get_proposed(user_projects)
+        project_dict[ProjectGroup('Active Projects', "active")] = Project.objects.get_active(user_projects)
+        project_dict[ProjectGroup('Completed Projects', "completed")] = Project.objects.get_completed(user_projects)
+
+        context["project_dict"] = project_dict
+        context["all_projects"] = list(chain(*(project_dict.values())))
+        context["role"] = "ambassador"
```
sameerav: Do we need role in the context? 

Also there seems to be a lot of repetition between the views - is there a way to consolidate them into a single template class that all three inherit from? and change what they need from that template class?

revolv/ambassador/views.py
```diff
@@ -2,21 +2,31 @@
...
+
+        # TODO: active proces processing
+        if len(context['all_projects']) > 0:
+            context['active_project'] = int(self.request.GET['active_project']) if 'active_project' in self.request.GET else context['all_projects'][0].id
```
sameerav: I think django templates have a way to get the length of a list in the html file! So it might not be necessary to pass it in as a variable if you're also passing in a list of projects somewhere

revolv/ambassador/views.py
noahsark769: We do need role, because in the case where the administrator looks at the admin dashboard, we need to know that they're looking as an ambassador rather than an admin.

As for the repetition, good point :) I'm not even actually sure if we need "all_projects" anymore, so I'll at least take that out.

revolv/static/auth.css
```diff
@@ -99,3 +90,31 @@ small.field-error {
...
+    background-color: transparent;
+    color: #f0ad4e;
+    border: 1px solid #f0ad4e; }
+    .submit-button.yellow:hover {
+      background-color: #f0ad4e;
+      color: white;
+      transition: background-color 0.2s;
+      transition: color 0.1s;
+      -webkit-transition: background-color 0.2s;
+      -webkit-transition: color 0.1s; }
```
sameerav: there's a lot of repeated styles in the .submit-button-gray and the .submit-button-yellow! maybe they could be consolidated?

revolv/static/auth.css
```diff
@@ -99,3 +90,31 @@ small.field-error {
...
+    .submit-button.gray:disabled {
+      opacity: 0.3; }
+  .submit-button.yellow {
+    cursor: pointer;
+    background-color: transparent;
+    color: #f0ad4e;
```
sameerav: this color (#f0ad4e) is used a ton! probably the other colors too

it would probably be good to make a variable for it!

revolv/templates/base/partials/dashboard_sidebar.html
```diff
@@ -0,0 +1,72 @@
...
+            etc, and values as lists of project model objects in the group. The project group names
+
+    The idea for the sidebar is that there is an area at the top where the user can switch between
+    different user types, if applicable, can see the projects which are relevant to their role, and can also
+    see management pages (for example admin accounting page) below. The point of this template is to
+    conditionally render all of these partials and things based on what type of user the current user is.
+{% endcomment %}
+{% spaceless %}
+<div class="role-select">
+    {% if role == "administrator" and user.revolvuserprofile.is_administrator %}
```
sameerav: why do we need role and user.revolvuserprofile.is_administrator?
sorry you're probably right but i'm just confused :(

sameerav:
screen shot 2015-04-14 at 10 14 37 pm

There's also a few problems when the screen gets smaller :(

the label on top "administrator" gets bigger than the margin it has, the buttons on the top right roll over

sameerav:
Shouldn't there also be a way to un-deny or un-accept a project?

revolv/static/auth.css
noahsark769: Yeah that's because they were compiled from the same mixin haha

revolv/ambassador/views.py
noahsark769: You are right about this! Just taking it out cause we don't actually use it anymore

revolv/static/auth.css
sameerav: oh oops this is a css not an scss my bad

revolv/static/auth.css
noahsark769: It is a variable in the SCSS haha, this is the compiled CSS

revolv/static/js/base/dashboard.js
noahsark769: Done!

revolv/templates/base/partials/dashboard_sidebar.html
noahsark769: Yeah, because the admin might not be viewing it as the sane role that s/he actually is

vivekraghuram:
@noahsark769 this looks sexy. i say ship it. caution is for the weak.

andylouisqin:
Noah this looks sick! Sorry I haven't been able to respond, but if I HAD to nitpick, it would be to change the line-height of the description to 24px so it's easier to read?:
screen shot 2015-04-15 at 12 43 50 am
Also iffy on the styling for each of these:
screen shot 2015-04-15 at 12 46 27 am
But besides those, this was really well done.

noahsark769:
@sameerav yeah, unfortunately mobile styles are nowhere to be seen at this point. Figured it was more important to get it out for the meeting tomorrow - will open an issue for that immediately after merging this.

As for un-denying and un-approving a project: you're right, we should probably have something there that can be a be-all end-all edit place for the project status. I'll open an issue for that too.

noahsark769:
@andylouisqin ohh you're right! I made the line-height 22px (24px seemed a tiny bit too large from my point of view) and I think it looks really nice. The role select links now underline on hover instead of being underlines by default.

noahsark769:
Thanks guys! Imma merge this in and open up a bunch more issues!

Issue #227, opened 2015-04-15
Write WebTests for testing dashboard flows


noahsark769:
Create project, propose project, etc

PR #248, opened 2015-04-27
removed the project review view and its corresponding redirects

sameerav:
removed the project review page class and its redirects and its url


noahsark769:
1. Codeship is unhappy, it seems you didn't take the review urls out of the tests
2. Will this still work? Like, will POSTing to view instead of review in the project controls partial actually complete/deny/etc the project? I feel like it won't?

noahsark769:
I think that you may have to leave in the view, but you can still take out the URLs, templates, etc

noahsark769:
In particular, we don't need `review_project.html`

noahsark769:
In any case, you should actually check that approve/deny/draft/post updates buttons and stuff still work. I wish we had tests for these, but those are in the pipeline in #227

sameerav:
oh you're right it doesn't work :(
i'll put the view and the url and change the form post urls back but add http_method_names = [u'post']
to the view so you can't see it

noahsark769:
Sounds great!

noahsark769:
@sameerav you should update this with master :)

noahsark769:
This is great, thanks!

PR #260, opened 2015-05-03
Move default categories from a fixture to a data migration

### LINKED ISSUE #257: Put categories in a data migration
Makes no sense to have them as just a fixture right now, because they should always be there!

noahsark769:
Fixes #257 

It doesn't make sense for us to have categories as a fixture, because some code depends on them always being in the database. So, I moved them to a data migration and removed the fixtures. Updated some tests to be more reliant on the fact that creating categories with default names may produce duplicate unique keys.

Could go for a lgtm from either @vivekraghuram or @erictshen


erictshen:
lgtm!

noahsark769:
```
                           |`-:_
  ,----....____            |    `+.
 (             ````----....|___   |
  \     _                      ````----....____
   \    _)                                     ```---.._
    \                                                   \ 
  )`.\  )`.   )`.   )`.   )`.   )`.   )`.   )`.   )`.   )`.   )hh
-'   `-'   `-'   `-'   `-'   `-'   `-'   `-'   `-'   `-'   `-'   `
```

PR #275, opened 2015-05-05
Change paypal charging back to always on

noahsark769:
I thought about only turning on charging for staging and prod, not local, but I decided against it in an effort to make all the environments as much the same as possible. @aykamko I'd like your opinion, do you think we should just never actually validate via paypal api on local?


noahsark769:
lol gunna merge this

PR #277, opened 2015-05-18
Updated the style of the admin email page

sameerav:
Basically:
-got rid of the buttons at the top because this page no longer serves as the dashboard
-made the buttons the revolv hover buttons in the dark blue color!

That's basically it!
Looks like this now: 
screen shot 2015-05-18 at 1 15 06 am

Looked like this before:
screen shot 2015-05-18 at 1 17 34 am

@noahsark769 could you review?


revolv/templates/administrator/email.html
```diff
@@ -1,42 +1,20 @@
 {% extends "base/base.html" %}
 {% load staticfiles %}
 
+{% block head %}
+    <link rel="stylesheet" type="text/css" href="{% static "email.css" %}" />
+    <link href='https://fonts.googleapis.com/css?family=Lato:300' rel='stylesheet' type='text/css'>
+{% endblock %}
+
 {% block title %}Home{% endblock %}
```
noahsark769: @sameerav I know this line wasn't changed, but would you mind updating it to have the actual correct title? haha

noahsark769:
I think this looks good! Probably we should eventually make this part of the dashboard, but super nice change for now. Can you update that one title? After that it should be good to go

noahsark769:
Nice, thanks Sameera!

PR #285, opened 2015-05-22
Fixed bug in typeahead address selector

### LINKED ISSUE #270: Address autocomplete and maps on project pages are totally broken
I think these two things are related

erictshen:
Fixes #270 

Bug was caused by the html div element for the map not actually being present in the html.

For projects that are currently buggy on staging, editing them and updating the address to a new one should fix things.

### DIFF
--- a/revolv/static/typeahead/typeahead.css
+++ b/revolv/static/typeahead/typeahead.css
@@ -40,7 +40,7 @@
 
 .tt-dropdown-menu {
   width: 100%;
-  margin-top: 12px;
+  margin-top: -15px;
   padding: 8px 0;
   background-color: #fff;
   border: 1px solid #ccc;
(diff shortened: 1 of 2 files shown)


erictshen:
screen shot 2015-05-21 at 5 32 19 pm

Screenshot of what the fixed page looks like!

noahsark769:
Wow nice @erictshen! Is it possible you can offset the CSS for the autocomplete popup so that it touches the text box? Other than that, this looks good

erictshen:
Updated it so the popup is closer to the textbox! Let me know if it looks good.

screen shot 2015-05-21 at 7 03 37 pm

noahsark769:
Looks great! Will merge when codeship passes!

PR #288, opened 2015-05-24
Project description/updates are now WYSIWYG editable

### LINKED ISSUE #256: Use CKEditor wysiwyg editor fields for project description, project update description fields
This will allow rich content editing! Yay!

### LINKED ISSUE #286: Get rid of templates/administrator/dashboard.html
This is a relic of the ancient past, don't need it anymore

noahsark769:
Fixes #256 and also quick fix: fixes #286 

This was actually a lot simpler than I thought it would be! django-ckeditor just dropped in pretty well. It can even do dynamic image uploads via S3!

screen shot 2015-05-24 at 12 15 25 pm
screen shot 2015-05-24 at 12 15 34 pm
screen shot 2015-05-24 at 12 16 00 pm
screen shot 2015-05-24 at 12 48 14 pm

Unfortunately, I don't think there is much to be done about making this work super efficiently on mobile: on tablet it for the most part works correctly, but it goes off the screen on mobile (though fortunately still usable). Seems like an alright compromise for now, we can try to efficiently mobilify it if we have time later.

Designating @sameerav as primary reviewer, @erictshen as secondary (if you have time). Need review from at least one of you!


revolv/static/project.scss
```diff
@@ -439,7 +439,7 @@ $gray-9: #999999;
             }
         }
 
-        .update-text {
+        .update-text p {
```
sameerav: couldn't you just delete this? i searched project.html but there are no items with class update-text that are of type p

revolv/static/project.scss
```diff
@@ -617,7 +617,7 @@ $map-height: 400px;
 
             }
 
-            .update-text {
+            .update-text p {
```
sameerav: same for this as above :)

sameerav:
looks good except for that little change in project.scss @noahsark769 !
ship it :)

revolv/static/project.scss
noahsark769: Nah, so the stuff rendered by ckeditor renders to html, and paragraph breaks are rendered as p tags. So then {{ project.description|safe }} will render p tags sometimes, so we need to have to style on those to make it look right!

Issue #291, opened 2015-05-26
Make sure we handle how RE-volv takes money from repayment correctly


noahsark769:
@aykamko 

So this is the breakdown of how repayments work with regard to RE-volv, the organization, taking money to run itself: say you have a project, with a 10,000 dollar funding goal, and it gets fully funded. Repayments will start coming in on a per month basis, and the interest rate and escalator rate for that project are set - that is, the community organization repays RE-volv what RE-volv gave it from the crowdfunding money, plus an interest rate, and the interest rate is escalated every year by an escalator rate.

Thus, when a repayment comes in, there is some portion of it which is return on the principle (the original money that was crowdfunded) and there is some portion of it which is due to the interest (and escalator) rates. 100% of the principle portion is made available for reinvestment. Of the interest portion, 60% is made available for reinvestment and 40% goes to RE-volv as an organization to cover maintenance costs, salaries, etc.

So what we really need to do, for the app, is probably make another payment type called `organizational_cost` or `revolv_cost` or something, which we don't take into account for reinvestments. Then, we can store a total of forty numbers (probably separate models I guess) associated with each project, where there are two number per year which represent the amount repaid to reinvestment (the principle + the 60% of the interest/escalator) and the amount repaid as organizational costs for RE-volv (the 40% of the interest/escalator).

Instead of calculating these numbers, Andreas said he would just prefer to enter them in manually - you only have to do it once per project, so this makes it easier on us. Then at the start and 15th of the month, if there are x dollars repaid to reinvestment and y dollars repaid to organizational costs for the current year of repayment for the project, we can just make new Payment objects representing x/2 dollars of type `repayment` and y/2 dollars of type `organizational_cost`, which will total out to x and y over the whole month.

What do you think of that @aykamko?

Issue #255, opened 2015-04-29
Add a project status: STAGED


noahsark769:
We should have a status for when a project is approved by the admin but not yet started (aka not accepting donations). Call it a project being "staged".

PR #295, opened 2015-05-28
Sameera accounting page

sameerav:
What I did:
1. Created the AdminAdjustment model and form and template, which allows Andreas to add expenses and income to the income statement.
2. Created the accounting page, which has buttons to filter with at the top (by start date, end date, and project) and a cash flow statement underneath. This page is accessible by both the administrators and donors.

What should still be done:
1. It would be super nice if the numbers on the page for each project donation, repayment, and reinvestment were CMSable!
2. We still haven't gotten final revolv earnings rate and payment processing fee numbers, so those still need to be entered.
3. For some reason, the values don't show up on the donor dashboard (I think this is an access thing but not sure? I'm figuring this out right now).
4. I still need to write tests (also working on that right now, but I'm pretty sure everything works from manually testing it. I'll keep doing that now).
5. Need to make the AccountingJSON view unviewable (not sure how to do this, working on it right now).

Screenshots! There are no project and project subheadings because no projects have been contributed to.

screen shot 2015-05-27 at 8 10 00 pm

screen shot 2015-05-27 at 8 14 26 pm

Screenshot of adjustment form:
screen shot 2015-05-27 at 8 12 15 pm

Could you review this @noahsark769 ?


noahsark769:
Looks great at first glance! Not sure if I'll be able to entirely review everything right now, but just a few things off the top of my head:

1. I think it's fine to have the accounting JSON view exposed if the accounting page is exposed - it has to be somewhere, and if we want to protect it we'd have to do some kind of weird stuff that would probably be breakable anyway, since everything is frontend. So I think it's fine to expose it!
2. We should make positive cash flow green and negative cash flow red :D
3. I guess the stuff that Andreas talked about at the skype meeting (aka the stuff in #291) will eventually replace the AdminAdjustment, because when #291 is implemented, we'll know the exact amounts per repayment that should go to the revolving fund and the exact amounts which should go to RE-volv - but I think it's worth keeping around anyway, just in case Andreas wants to adjust the totals on this page. So, nice!

revolv/administrator/urls.py
```diff
@@ -10,4 +13,7 @@
     url(r'^$', is_administrator(AdministratorDashboardView.as_view()), name='dashboard'),
     url(r'^email$', is_administrator(AdministratorEmailView.as_view()), name='email'),
     url(r'^email/csv$', admin_email_csv_download, name='emailcsv'),
+    url(r'^accounting$', is_administrator(AdministratorAccountingView.as_view()), name='accounting'),
```
noahsark769: The `is_administrator` here might be why it's not visible to a donor?

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+	placeElements(data, 'cash_out', 'reinvestments', placeAdminInvestmentElement);
+	placeElements(data, 'cash_out', 'adjustments', placeCashOutAdjustmentElement);
+}
+
+/**
+ * Cleanes up the String to be JQuery friendly.
+ * @param {String} text - The text to be cleaned up to be made JQuery friendly.
+ */
+function cleanUpString(text) {
+	text = text.split(' ').join('-').split(';').join('sm').split(":").join('co').split("'").join("ap").split("!").join('ex');
```
noahsark769: OH MY GOOOOOOD

revolv/administrator/forms.py
```diff
@@ -0,0 +1,27 @@
+from django import forms
+from revolv.payments.models import AdminAdjustment
+from revolv.project.models import Project
+
+class AdjustmentForm(forms.ModelForm):
```
noahsark769: Just a thought, would it be easier to use [modelform_factory](https://docs.djangoproject.com/en/1.7/topics/forms/modelforms/#modelform-factory-function)?

revolv/administrator/tests.py
```diff
@@ -1,3 +0,0 @@
-from django.test import TestCase
```
noahsark769: Nice, this needed to be cleaned up

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class AdministratorAccountingView(UserDataMixin, TemplateView):
+    """View of the administrator accounting page.
```
noahsark769: Can you put a newline between the `"""` and the `View` to match the formatting of other docstrings?

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+    """
+    template_name = 'administrator/accounting.html'
+
+    def clean_project_name(self, name):
```
noahsark769: This could use a docstring...but that replace is kind of ugly so I be there might be a better way to do it haha

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        return context
+
+class AccountingJSONView(View):
+    """
+    A view that returns a JsonResponse of all relevant accounting data.
+    This is queried by an AJAX request when creating the Accounting page.
+
+    The data is roughly structured as shown below:
+    data = {
+        'cash_in': {},
```
noahsark769: You should provide structures for what the inner objects look like also, that would make it very understandable!

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+    """
+
+    start_date = datetime.datetime.strptime('2015-01-01', '%Y-%m-%d')
+    end_date = datetime.datetime.now()
+    project_filter = Project.objects.all()
+
+    project_dict = {}
+
+
+    """
```
noahsark769: In this file, there are a couple things that need to be fixed everywhere:

1. The docstrings should go directly below the function definitions, not above
2. All python code should be snake_case (not camelCase) to match all the other python code

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        'clean_project_names': [],
+        }
+    """
+
+    start_date = datetime.datetime.strptime('2015-01-01', '%Y-%m-%d')
+    end_date = datetime.datetime.now()
+    project_filter = Project.objects.all()
+
+    project_dict = {}
+
```
noahsark769: Only need one blank line here, not two :)

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+    beginning of time to self.start_date
+
+    :return:
+        Returns a number that represents initial cash balance.
+    """
+    def calculateCashBalanceFromStart(self):
+        # THESE ARE DUMMY VARIABLES - USE THE REVOLV EARNINGS MODEL INSTEAD
+        payment_service_fee_rate = 0.01
+        revolv_earnings_rate = 0.07
+        allDonations = sum([payment.amount for payment in Payment.objects.donations(end_date=self.start_date)])
```
noahsark769: Would it be better to use django's Sum aggreator? ([documentation](https://docs.djangoproject.com/en/1.7/topics/db/aggregation/))

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        cashInAdjustments = sum([payment.amount for payment in AdminAdjustment.objects.adjustments(end_date=self.start_date, cash_type='cash_in')])
+        cashOutAdjustments = sum([payment.amount for payment in AdminAdjustment.objects.adjustments(end_date=self.start_date, cash_type='cash_out')])
+
+        cash_in = (allDonations * (1 - payment_service_fee_rate)) + (allRepayments * (1 - revolv_earnings_rate)) + cashInAdjustments
+        cash_out  = cashOutAdjustments + allReinvestments
+        net = cash_in - cash_out
+        return net
+    
+    """
+    A helper function that cleans a project name to make it JQuery
```
noahsark769: What do you mean by "jQuery friendly"? It might just be better to URL encode it ([python urlencode](https://docs.python.org/2/library/urllib.html#urllib.urlencode), [javascript decodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent))

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        name: A string to be cleaned
+
+    :return:
+        Returns a cleaned up version of name.
+    """
+    def clean_project_name(self, name):
+        name = name.replace(' ', '-').replace(';', 'sm').replace(':', 'co').replace("'", 'ap').replace('!', 'ex')
+        return name
+    
+    def make_project_dict(self):
```
noahsark769: Might be good to memoize this (aka only query the `Project.objects.all()` if you haven't called the function already), because querying all Projects may be slow if this is called a lot of times

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        self.project_dict = {}
+        for proj in Project.objects.all():
+            name = proj.title
+            self.project_dict[self.clean_project_name(name)] = name
+    
+    """
+    Updates class variables such as start_date, end_date, and project_filter
```
noahsark769: Nitpick, they're actually instance variables, not class variables

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+                self.start_date = datetime.datetime.strptime(keys['start_date'], '%Y-%m-%d')
+            elif 'end_date' in keys:
+                self.end_date = datetime.datetime.strptime(keys['end_date'], '%Y-%m-%d')
+            elif 'project_choice' in keys:
+                if keys['project_choice'] == 'all':
+                    self.project_filter = Project.objects.all()
+                else:
+                    real_name = self.project_dict[keys['project_choice']]
+                    self.project_filter = Project.objects.filter(title=real_name)
+        except ValueError:
```
noahsark769: I feel like this deserves a comment saying when it would be raised and why it's safe to ignore it

revolv/administrator/urls.py
sameerav: nah, there's a separate url in the donor dashboard for donor access that doesn't check is_administrator

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+
+    :return:
+        Returns a properly set up dictionary.
+    """
+    def setUpFilteredDict(self):
+        filtered_dict = {'cash_in': {'repayments': {}, 'adjustments': {}, 'donations': {}}, 'cash_out': {'reinvestments': {}, 'adjustments': {}}}
+        
+        filtered_dict['date_info'] = {}
+        filtered_dict['cash_balances'] = {}
+        date_info = filtered_dict['date_info']
```
noahsark769: It would be cleaner to set `date_info` to an empty dict at the beginning and then set the filled in dict to be part of the filtered_dict at the end

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        filtered_dict['project_names'] = []
+        filtered_dict['clean_project_names'] = []
+        clean_project_names = filtered_dict['clean_project_names']
+        project_names = filtered_dict['project_names']
+        for proj in self.project_filter:
+            if (self.hasPayments(proj)):
+                project_names.append(proj.title)
+                clean_project_names.append(self.clean_project_name(proj.title))
+
+        filtered_dict['cash_in_adjustment_names'] = list(set([adj.name for adj in AdminAdjustment.objects.adjustments(start_date=self.start_date, end_date=self.end_date, cash_type='cash_in')]))
```
noahsark769: Instead of doing `set`, you can just do a `.distinct()` on the queryset I believe

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        return (donations + reinvestments + repayments) != 0
+
+    """
+    A helper function that returns a date one month ahead of the date
+    passed in.
+
+    :return:
+        A Python datetime.datetime object with a time value one month ahead of date.
+    """
+    def moveOneMonthForward(self, date):
```
noahsark769: This one should definitely be in a utility somewhere, not in the View. In fact, what about putting all this computation stuff in a utility file, like a `utils.AccountingAggregator` class or something like that? There is a lot of logic in the view right now which would probably be easier to understand if it was extracted

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        clean_project_names = context['clean_project_names']
+        for p in context['project_names']:
+            clean_project_names[self.clean_project_name(p)] = p
+
+        return context
+
+class AccountingJSONView(View):
+    """
+    A view that returns a JsonResponse of all relevant accounting data.
+    This is queried by an AJAX request when creating the Accounting page.
```
noahsark769: Could we maybe talk here about how the accounting info is calculated and how the AdminAdjustments play into it?

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+
+    """
+    Sets up subdictionaries of each Payment object model for a certain date.
+
+    :args:
+        filtered_dict: The JSON serializable dictionary to be modified.
+        date: The current date period we are iterating over.
+    """
+    def makePaymentTypeKeys(self, filtered_dict, date):
+        filtered_dict['cash_in']['repayments'][date] = {}
```
noahsark769: The next fives lines could be done in a two line for loop :)

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+    
+    """
+    Sets up subdictionaries of each Payment object model for a certain date and project.
+
+    :args:
+        filtered_dict: The JSON serializable dictionary to be modified.
+        date: The current date period we are iterating over.
+        title: The title of the project.
+    """
+    def makeProjectKeys(self, filtered_dict, date, title):
```
noahsark769: This could also be a for loop

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+    Creates and fills in the 'adjustment' subdictionary of 'cash_in' and 'cash_out'
+    within the JSON serializable dictionary.
+
+    :args:
+        cash: A subdictionary of the JSON serializable dictionary filtered_dict.
+        cash_balance: The cash balance at the beginning of this period.
+        net_cash_in: Net inflow of cash during this period.
+        net_cash_out: Net outflow of cash during this period.
+    """
+    def setTimePeriodAdjustments(self, filtered_dict, date, temp_start_date, temp_end_date):
```
noahsark769: This function should definitely be more heavily commented, it's very tough to tell what's going on

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        temp_start_date: The start date of this time period.
+        temp_end_date: The end date of this time period.
+        revolv_earnings_rate: A float from 0-1 representing what percentage of repayments RE-volv takes as earnings.
+    """
+    def setRepaymentTableValues(self, filtered_dict, date, proj, temp_start_date, temp_end_date, revolv_earnings_rate):
+        current_dict = filtered_dict['cash_in']['repayments'][date][proj.title]
+        total = sum([repayment.amount for repayment in AdminRepayment.objects.repayments(project=proj, start_date=temp_start_date, end_date=temp_end_date)])
+        current_dict['total'] = total
+        current_dict['revolv_earnings'] = total * revolv_earnings_rate
+        current_dict['retained_ROI'] = total * (1 - revolv_earnings_rate)
```
noahsark769: Doesn't ROI stand for "rate of interest"? Also shouldn't this value be just "retained earnings" instead?

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+                net_cash_in += cash_in_donation_date[proj.title]['retained_donations']
+                net_cash_out += cash_out_reinvestment_date[proj.title]['total']
+            
+            self.setCashBalances(cash, cash_balance, net_cash_in, net_cash_out)
+            cash_balance += net_cash_in - net_cash_out
+
+            temp_start_date = temp_end_date
+            temp_end_date = self.moveOneMonthForward(temp_end_date)
+        
+        return JsonResponse(filtered_dict)
```
noahsark769: Imagine if we could just say `return JsonResponse(new AccountingAggregator().aggregate())`

revolv/administrator/views.py
```diff
@@ -29,6 +35,423 @@ def get_context_data(self, **kwargs):
...
+        revolv_earnings_rate: A float from 0-1 representing what percentage of repayments RE-volv takes as earnings.
+        payment_service_fee_rate: A float from 0-1 representing what percentage of donations that are taken by payment service fees.
+    """
+    def setAlTableValues(self, filtered_dict, date, proj, temp_start_date, temp_end_date, revolv_earnings_rate, payment_service_fee_rate):
+        self.setRepaymentTableValues(filtered_dict, date, proj, temp_start_date, temp_end_date, revolv_earnings_rate)
+        self.setDonationTableValues(filtered_dict, date, proj, temp_start_date, temp_end_date, payment_service_fee_rate)
+        self.setReinvestmentTableValues(filtered_dict, date, proj, temp_start_date, temp_end_date)
+
+    """
+    Creates and returns a JsonResponse.
```
noahsark769: This function does a bit more than that haha, it's like 80 lines. Could definitely use some more comments

revolv/donor/urls.py
```diff
@@ -1,8 +1,10 @@
 from django.conf.urls import patterns, url
 from django.contrib.auth.decorators import login_required
 from revolv.donor.views import DonorDashboardView
+from revolv.administrator.views import AdministratorAccountingView
 
 urlpatterns = patterns(
     '',
     url(r'^$', login_required(DonorDashboardView.as_view()), name='dashboard'),
+    url(r'^accounting$', AdministratorAccountingView.as_view(), name='accounting'),
```
noahsark769: Wasn't this just in another urls.py?

revolv/payments/models.py
```diff
@@ -9,7 +11,7 @@ class AdminRepaymentManager(models.Manager):
     Manager for AdminRepayment.
     """
 
-    def repayments(self, admin=None, project=None, queryset=None):
+    def repayments(self, admin=None, project=None, queryset=None, start_date=datetime.datetime.min, end_date=datetime.datetime.max):
```
noahsark769: May be better to default these to None? That way if they're not provided, the query doesn't care about time restraints at all

revolv/payments/models.py
```diff
@@ -253,12 +261,55 @@ class RepaymentFragment(models.Model):
...
+        if cash_type:
+            queryset = queryset.filter(cash_type=cash_type)
+        
+        return queryset
+
+class AdminAdjustment(models.Model):
+    """
+    Abstraction indicating one RE-volv adjustment to the accounting.
+    
+    This model exists for accounting reasons.
```
noahsark769: Can you expand on this? haha

revolv/payments/models.py
```diff
@@ -253,12 +261,55 @@ class RepaymentFragment(models.Model):
...
+    """
+    amount = models.FloatField("How many dollars is this transaction?")
+    name = models.CharField(
+            "What is the name of this expense or income?",
```
noahsark769: These should be indented down one level

noahsark769: Oh no they shouldn't nevermind. I think the closing paren should be on a new line and de-indented though

revolv/payments/models.py
```diff
@@ -253,12 +261,55 @@ class RepaymentFragment(models.Model):
...
+    name = models.CharField(
+            "What is the name of this expense or income?", 
+            max_length=100)
+
+    CASH_TYPE_CHOICES = (
+        ('cash_in', 'Income'),
+        ('cash_out', 'Expense'),
+    )
+    cash_type = models.CharField(
+                "Is this income or an expense?",
```
noahsark769: THIS should be indented down a level

revolv/payments/models.py
```diff
@@ -253,12 +261,55 @@ class RepaymentFragment(models.Model):
...
+        ('cash_out', 'Expense'),
+    )
+    cash_type = models.CharField(
+                "Is this income or an expense?", 
+                choices=CASH_TYPE_CHOICES, 
+                max_length=10, 
+                default='cash_out')
+
+    created_at = models.DateField(
+                    "What time period should this transaction be recorded under?",
```
noahsark769: THIS should be indented down TWO levels lol

revolv/static/accounting.scss
```diff
@@ -0,0 +1,115 @@
+body {
+	background-color: white;
```
noahsark769: BUT WHY TABS THO

revolv/static/accounting.scss
```diff
@@ -0,0 +1,115 @@
...
+	padding: 0;
+}
+
+.heading {
+	.revolv-header-text {
+		font-size: 1.3rem;
+		font-weight: 800;
+		margin-top: 30px;
+	}
+	.filter_form {
```
noahsark769: Classes should have the dashed-snake-case form

revolv/static/accounting.scss
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		.cash-out-head-row {
+			th {
+				padding-top: 25px;
+				align: bottom;
+			}
+		}
+
+		.level-0 {
```
noahsark769: I think these could be more descriptive?

revolv/static/adjustment.scss
```diff
@@ -0,0 +1,25 @@
+@import "utils";
+@import "revolv-settings";
+
+.save-button {
+	@include revolv-hover-button($revolv-dark-blue, $revolv-dark-blue);
+	width: 97%;
```
noahsark769: Why 97%? Shouldn't it be 100%?

revolv/static/adjustment.scss
```diff
@@ -0,0 +1,25 @@
...
+
+.container {
+	margin-top: 50px;
+}
+
+body {
+	position: inline;
+}
+
+#id_amount {
```
noahsark769: In general you shouldn't target ids in css, use class names instead :)

revolv/static/adjustment.scss
```diff
@@ -0,0 +1,25 @@
...
+	position: absolute;
+	margin-top: -45px;
```
noahsark769: Also, wat, this seems kind of wonky, is there a reason it needs to have negative margin? If it's position: absolute shouldn't you just define top and left?

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
+$(document).ready(function(){
+	//sets up initial values when the page is loaded
```
noahsark769: but why TABS DOE

noahsark769: Also, one space after `//` :D

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
+$(document).ready(function(){
+	//sets up initial values when the page is loaded	
+	makeAJAXcall({});
+})
+
+$(".selector").change(function(){
+	// makes an AJAX call to the accountingJSON view when the selectors are changed
+	
+	$(".added").remove();
+	var changed_selector = $(this);
```
noahsark769: javascript should always be camelCase

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+	var changed_data = {};
+
+	if (changed_selector.hasClass("start_selector")) {
+		changed_data['start_date'] = changed_selector[0].value;
```
noahsark769: Shouldn't you do `changed_selector.val()`?

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+	} else if (changed_selector.hasClass("end_selector")) {
+		changed_data['end_date'] = changed_selector[0].value;
+	} else {
+		changed_data['project_choice'] = changed_selector[0].value;
+	}
+	console.log(changed_data)
```
noahsark769: Have you run pre-commit install? jshint should have cought this console.log :(

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+	makeAJAXcall(changed_data);
+});
+
+/**
+ * Makes the AJAX Call
+ * @param {Object} changed_data - An object with either 0 or 1 keys that maps a selector to its new value.
+ */
+function makeAJAXcall(changed_data) {
```
noahsark769: Random nitpick: the c in "call" should be capitalized

revolv/static/js/base/accounting.js
noahsark769: Yeah I bet you should just URL decode this and URL encode it on the backend instead haha

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+}
+
+/**
+ * Places a donation element on the page.
+ * @param {Object} data_to_insert - A dictionary with values to insert,
+ * @param {String} project_name - The name of the project we are inserting under.
+ * @param {String} time_key - The time period in which this was inserted.
+ */
+function placeDonationElement(data_to_insert, project_name, time_key) {
+	var total = data_to_insert['total'];
```
noahsark769: camelCase

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+	var payment_service_fees = data_to_insert['payment_service_fees'];
+	var retained_donations = data_to_insert['retained_donations'];
+	var className  = '.project-' + cleanUpString(project_name);
+	$('.total-donation').filter(className).append('<td class="added">' + String(total) + "</td>");
```
noahsark769: Casting to string like this is unneeded in javascript, since you're adding it to a string is will get automatically converted. Also isn't it a string already anyway?

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+	$('.payment-service-fees').filter(className).append('<td class="added">' + String(payment_service_fees) + "</td>");
+	$('.retained-donations').filter(className).append('<td class="added">' + String(retained_donations) + "</td>");
+}
+
+/**
+ * Places an AdminRepayment element on the page.
+ * @param {Object} data_to_insert - A dictionary with values to insert,
```
noahsark769: camelCase

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+ * @param {String} project - The name of the project we are inserting under.
+ * @param {String} time - The time period in which this was inserted.
+ */
+function placeROIElement(data_to_insert, project, time) {
```
noahsark769: camelCase

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+var allCashInTransactions = {};
+function placeCashInAdjustmentElement(data_to_insert, time) {
+	var total = data_to_insert['total'];
+	var transactions = data_to_insert['transactions'];
+
+	var allTransactionsKeys = Object.keys(allCashInTransactions);
+	for (var i = 0; i < allTransactionsKeys.length; i++) {
+		allCashInTransactions[allTransactionsKeys[i]] = 0;
+	}
+	//insert all the transactions
```
noahsark769: One space after // :D

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+	$('.cash-in-adjustment-total').append('<td class="added">' + total + "</td>");
+}
+
+/**
+ * Places a cash out AdminAdjustment element on the page.
+ * @param {Object} data_to_insert - A dictionary with values to insert.
+ * @param {String} time - The time period in which this was inserted.
+ */
+var allCashOutTransactions = {};
+function placeCashOutAdjustmentElement(data_to_insert, time) {
```
noahsark769: Isn't this almost exactly the same as the last function? I.e. couldn't you just extract this functionality?

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+		$('.transaction').filter('.' + name).append('<td class="added">' + amount + "</td>");
+	}
+
+	$('.cash-out-adjustment-total').append('<td class="added">' + total + "</td>");
+}
+
+/**
+ * Iterates through the data and inserts values onto the page.
+ * @param {Object} data - A dictionary of all the data
+ * @param {String} cash_in_or_cash_out - 'cash_in' for cash in, and 'cash_out' for cash out.
```
noahsark769: Do you think it would be better to just have a boolean where true -> cash in, false -> cash out?

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+ */
+function placeElements(data, cash_in_or_cash_out, subsection, insert_func) {
+	if (subsection == 'adjustments') {
```
noahsark769: ===

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+		var super_dict = data[cash_in_or_cash_out][subsection]
+		var time_keys = Object.keys(super_dict);
+		time_keys.sort();
+		for (var i = 0; i < time_keys.length; i++) {
+			var key = time_keys[i];
+			var transactionsAndTotal = super_dict[key];
+			insert_func(transactionsAndTotal, key);
+		}
+	} else {
+		var super_dict = data[cash_in_or_cash_out][subsection]
```
noahsark769: camelCase

revolv/static/js/base/accounting.js
noahsark769: Also you need docs for subsection and insert_func

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+		}
+	}
+}
+
+/**
+ * Sets the overal totals and the cash flow values.
+ * @param {Object} data - A dictionary with values to insert.
+ */
+function setTotalsAndCashFlow(data) {
+	//create the total row
```
noahsark769: one space after // lol

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,388 @@
...
+	//iterate through and create all the cash positions at each time
+	$('.cash-in-body').append('<tr class="added level-0 total-cash-in"><td class="level-0 added">Total Cash In</td><tr>');
```
noahsark769: It looks like you probably don't need to append these, you can just render them by default and fill in their contents?

revolv/templates/administrator/accounting.html
```diff
@@ -0,0 +1,83 @@
+{% extends "base/base.html" %}
+{% load staticfiles %}
+{% load widget_tweaks %}
+
+{% block head %}
+    <link rel="stylesheet" type="text/css" href="{% static "accounting.css" %}" />
+    <link href='https://fonts.googleapis.com/css?family=Lato:300,500,800' rel='stylesheet' type='text/css'/>
+    <script type="text/javascript">
+		window.JSONURL = "{{ jsonurl }}";
```
noahsark769: BUT WHY TABS THOOOOOOO

revolv/templates/administrator/accounting.html
```diff
@@ -0,0 +1,83 @@
...
+					<option value={{ cleantitle }}>{{ title }}</option>
+				{% endfor %}
+			</select>
+			{% if is_administrator %}
+				<div class="expense-link"><a class="link" href="{% url "administrator:adjustment" %}">Add Expense/Adjustment<a/></div>
+			{% endif %}
+		</div>
+	</form>
+</div>
+
```
noahsark769: No newline needed here, I think

revolv/templates/administrator/adjustment.html
```diff
@@ -0,0 +1,37 @@
+{% extends "base/base.html" %}
+{% load staticfiles %}
+{% load widget_tweaks %}
+
+{% block head %}
+<link rel="stylesheet" href="{% static "adjustment.css" %}" />
```
noahsark769: Indent

revolv/templates/administrator/adjustment.html
```diff
@@ -0,0 +1,37 @@
...
+<link href='http://fonts.googleapis.com/css?family=Lato:300,500,800' rel='stylesheet' type='text/css'>
+
```
noahsark769: Delete this line :D

revolv/administrator/urls.py
noahsark769: Oh, yeah I see

noahsark769:
Okay, so, all of this javascript is INCREDIBLY complicated. It leads me to think that we might want to pursure a better way to render the page.

What we could do: instead of rendering to JSON in the AJAX view, we render to a partial template instead, where we can use template functions to iterate through all of this stuff instead of having to parse it and place it with js. This would only be a partial html document, with no styles or `<head>` or anything - but say it accepts GET parameters like `/accounting/partial/?start-date=2015-08-15&end-date=2015-08-17` and renders html for those dates which we can then load in with [jQuery's load() method](http://api.jquery.com/load/). With this way, you'd still get to keep all the backend logic you wrote, and all of the javascript logic would basically become template logic.

What do you think about that? Another option is to do the templating on the client side with something like [underscore templates](http://www.bennadel.com/blog/2411-using-underscore-js-templates-to-render-html-partials.htm). In this case, we still send the data as JSON in the AJAX view, but we have a template stored in some element on the client that we render to every time we get the data (basically moving all the logic that "places" elements to be owned by the underscore templating instead). Not sure, but I do know that javascript like can get to be practically un-understandable (trust me, I've written some, haha).

noahsark769:
On the bright side, when this gets merged, we'll probably be up to almost 970 commits!

revolv/static/js/base/accounting.js
sameerav: @noahsark769 wait i dont think URL encoding will work because some characters in the ascii character set aren't allowed in css class names

revolv/administrator/forms.py
sameerav: probably oops!

revolv/static/js/base/accounting.js
noahsark769: Hm, I don't understand why the CSS class names would impact the way you do encoding, can you explain that?

sameerav: URL encoding keeps all ascii characters except the space, but certain ascii characters(like exclamation marks and colons and semicolons and apostrophes) and the reason I'm using modified strings is so those names can go in the classnames of their respective html elements...
but now that I think about it it might just be better to use project ids

revolv/administrator/views.py
sameerav: ROI is actually 'return on investment' so I think it should be okay! I didn't want to put retained earnings because I think that's easily confused with revolv_earnings, but i'm not sure

do you think that would be more clear?

revolv/administrator/views.py
sameerav: screen shot 2015-05-28 at 7 53 18 pm
I AGREE THIS IS SO MUCH BETTER
this is the light

revolv/donor/urls.py
sameerav: this is the donor view!

revolv/static/js/base/accounting.js
noahsark769: Yeah I think project ids would be better haha

revolv/administrator/views.py
noahsark769: Oh whoops, I hate acronyms. What about `retained_returns`? Or I guess `retained_ROI` is fine haha

revolv/administrator/views.py
noahsark769: YAY

revolv/donor/urls.py
noahsark769: Shouldn't there just be one though?

noahsark769: Like we can put it in the main urls.py if needed (which might make sense since it's not specific to donors or admins)

revolv/static/adjustment.scss
sameerav: nah there's padding on the side of the columns

noahsark769: Oh, I think it would be better to do width: 100% and box-sizing: border-box in order to make it more precise

revolv/donor/urls.py
sameerav: yeah sure
what do you think the url would be?

sameerav:
I think this is done! @noahsark769 could you review?
screen shot 2015-06-01 at 5 08 03 pm
screenshot

noahsark769:
Oh oops @sameerav yup I'll review this tomorrow morning

noahsark769:
Style things: I think "statement of cash flows" and "RE-volv accounting information" both need more margin above them

revolv/administrator/forms.py
```diff
@@ -0,0 +1,7 @@
+from django import forms
+from revolv.payments.models import AdminAdjustment
+from revolv.project.models import Project
+from django.forms.models import modelform_factory
+
+
+AdjustmentForm = modelform_factory(AdminAdjustment)
```
noahsark769: You should add `exclude=[]` so the logger doesn't complain at us about being removed in django 1.8

revolv/administrator/views.py
```diff
@@ -1,11 +1,14 @@
...
-from django.views.generic import TemplateView
+from django.views.generic import TemplateView, CreateView
 from revolv.base.models import RevolvUserProfile
 from revolv.base.users import UserDataMixin
 from revolv.base.views import BaseStaffDashboardView
-
+from revolv.payments.models import AdminAdjustment
+from revolv.administrator.forms import AdjustmentForm
+from revolv.project.models import Project
+from django.core.urlresolvers import reverse
```
noahsark769: Nitpick: this should go up with the other django imports, not with the revolv imports

revolv/administrator/views.py
```diff
@@ -28,6 +31,18 @@ def get_context_data(self, **kwargs):
...
+    
+    model = AdminAdjustment
+    template_name = 'administrator/adjustment.html'
+
+    def get_success_url(self):
+        return reverse('administrator:accounting')
+
+    def get_form_kwargs(self):
+        kwargs = super(AdministratorAddAdjustmentView, self).get_form_kwargs()
+        self.form = AdjustmentForm()
```
noahsark769: I'm reading the source of CreateView and I can't seem to find why you would need to set self.form here - why?

revolv/donor/urls.py
```diff
@@ -1,8 +1,10 @@
 from django.conf.urls import patterns, url
 from django.contrib.auth.decorators import login_required
-from revolv.donor.views import DonorDashboardView
+from revolv.donor.views import DonorDashboardView, AccountingView, AccountingPartialView
 
 urlpatterns = patterns(
     '',
     url(r'^$', login_required(DonorDashboardView.as_view()), name='dashboard'),
+    url(r'^accounting$', AccountingView.as_view(), name='accounting'),
+    url(r'^accounting/partial$', AccountingPartialView.as_view(), name='accountingPartial')
```
noahsark769: accountingPartial should be accounting_partial I think

revolv/donor/views.py
```diff
@@ -30,3 +34,78 @@ def get_context_data(self, **kwargs):
...
+    def get_context_data(self, **kwargs):
+        context = super(AccountingView, self).get_context_data(**kwargs)
+
+        context['project_names'] = [p.title for p in Project.objects.all()]
+        context['admin_repayments'] = AdminRepayment.objects.repayments()
+        context['admin_reinvestments'] = AdminReinvestment.objects.reinvestments()
+        context['donations'] = Payment.objects.donations()
+        context['jsonurl'] = unicode(str(reverse('donor:accountingPartial')))
+
+        context['clean_project_names'] = {}
```
noahsark769: I'm still not sure why you need clean project names to be class names in html. Can't you just called the classes `project-{{project.pk}}`?

revolv/donor/views.py
```diff
@@ -30,3 +34,78 @@ def get_context_data(self, **kwargs):
...
+        for p in context['project_names']:
+            clean_name = AccountingAggregator().clean_project_name(p)
+            clean_project_names[clean_name] = p
+
+        return context
+
+class AccountingPartialView(TemplateView):
+    """
+    A view that returns partial HTML of all relevant accounting data.
+    This is queried by an AJAX request when creating the Accounting page.
```
noahsark769: You should say that not only is it queried when creating the accounting page, but when requesting data for new dates as well.

revolv/donor/views.py
```diff
@@ -30,3 +34,78 @@ def get_context_data(self, **kwargs):
...
+    """
+
+    start_date = datetime.datetime.strptime('2015-01-01', '%Y-%m-%d')
+    end_date = datetime.datetime.now()
+    project_filter = Project.objects.all()
+    allProjects = Project.objects.all()
```
noahsark769: should be all_projects

revolv/donor/views.py
```diff
@@ -30,3 +34,78 @@ def get_context_data(self, **kwargs):
...
+        elif 'project_choice' in request:
+            if request['project_choice'] == 'all':
+                self.project_filter = Project.objects.all()
+            else:
+                project_dict = {}
+                for proj in self.allProjects:
+                    name = proj.title
+                    project_dict[AccountingAggregator().clean_project_name(name)] = name
+                real_name = project_dict[request['project_choice']]
+                self.project_filter = Project.objects.filter(title=real_name)
```
noahsark769: I think it would be much nicer to have the `project_choice` be a project id instead, then you could just do a `Project.objects.get` here instead of doing all this cleaning/filtering

revolv/donor/views.py
```diff
@@ -30,3 +34,78 @@ def get_context_data(self, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        context = super(AccountingPartialView, self).get_context_data(**kwargs)
+        revolv_earnings_rate = 0.07
+        payment_service_fee_rate = 0.01
```
noahsark769: According to [https://developer.paypal.com/docs/faq/](https://developer.paypal.com/docs/faq/), "For Purchase Payments (which include payments processed using the REST APIs or PayPal Mobile SDK), PayPal U.S. charges 2.9% or less, plus $0.30 per transaction." Can we account for that here? We should make that a constant somewhere, because according to [https://www.paypal.com/webapps/mpp/paypal-fees](https://www.paypal.com/webapps/mpp/paypal-fees), eligible nonprofits can apply for as low as 2.2%, so the rate may change.

revolv/donor/views.py
```diff
@@ -30,3 +34,78 @@ def get_context_data(self, **kwargs):
...
+    This is queried by an AJAX request when creating the Accounting page.
+    """
+
+    start_date = datetime.datetime.strptime('2015-01-01', '%Y-%m-%d')
+    end_date = datetime.datetime.now()
+    project_filter = Project.objects.all()
+    allProjects = Project.objects.all()
+    template_name = 'base/partials/cash_flow_table.html'
+
+    def updateFilterVariables(self, request):
```
noahsark769: again this should be update_filter_variables

revolv/payments/migrations/0031_auto_20150602_0011.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
noahsark769: There are a ridiculous amount of migration files here, and running migrations takes long enough as it is. Do you think you can delete the new ones then run a `makemigrations` again so that there's only one payments migration file merged with this PR?

revolv/payments/models.py
```diff
@@ -19,6 +21,13 @@ def repayments(self, admin=None, project=None, queryset=None):
             queryset = queryset.filter(admin=admin).order_by('created_at')
         if project:
             queryset = queryset.filter(project=project).order_by('created_at')
+        if start_date and not end_date:
```
noahsark769: Not sure why these two if statements need to be here. Can't you just filter on end date if you see an end date and filter on start date if you see a start date?

revolv/payments/models.py
```diff
@@ -78,6 +87,13 @@ def reinvestments(self, admin=None, project=None, queryset=None):
...
```
noahsark769: Same thing here. In fact, this code is repeated exactly from above (and again down on line 275). Can you extract it into a utility?

revolv/payments/models.py
```diff
@@ -270,9 +339,16 @@ def payments(self, user=None, entrant=None, project=None, queryset=None):
             queryset = queryset.filter(entrant=entrant).order_by('created_at')
         if project:
             queryset = queryset.filter(project=project).order_by('created_at')
+        if start_date and not end_date:
```
noahsark769: Here's another place which is repeated from above

revolv/static/accounting.scss
```diff
@@ -0,0 +1,115 @@
...
+        margin-bottom: 100px;
+
+        td, th {
+            padding: 3px;
+            font-size: 0.9rem;
+            font-weight: 300;
+        }
+
+        .payment-service-fees, .revolv-earnings {
+            color: red;
```
noahsark769: This should be the red from revolv-settings - also, it doesn't seem to appear in the screenshot

revolv/static/adjustment.scss
```diff
@@ -0,0 +1,25 @@
+@import "utils";
+@import "revolv-settings";
+
+.save-button {
+    @include revolv-hover-button($revolv-dark-blue, $revolv-dark-blue);
+    width: 97%;
```
noahsark769: Again, you can make this box-sizing: border-box and width: 100% to deal with the padding: making it 97% is a bit hacky

revolv/payments/utils.py
```diff
@@ -1,6 +1,398 @@
...
+    """
+    An aggregator class that does much of the computation for the AdministratorAccountingView
+    view in revolv.administrator.views.
+    This class contains a bunch of helper functions and one main function (aggregate) which returns 
+    a dictionary of relevant accounting data, filtered by instance variables representing the start date,
+    end date, and project filters.
+    """
+
+    def __init__(self, start_date=None, end_date=None, project_filter=None, revolv_earnings_rate=0, payment_service_fee_rate=0):
+        # to prevent uneccessary operations
```
noahsark769: If you use project ids, then you won't have to clean the project names, which means you'll always want to instantiate this with all the arguments, which means you won't need this if clause. Also, couldn't you technically want to instantiate an aggregator with no start date? That case would be a big bug, because nothing else would get instantiated even if you passed it in.

revolv/payments/utils.py
```diff
@@ -1,6 +1,398 @@
...
+        :return:
+            Returns a number that represents initial cash balance.
+        """
+        all_donations = Payment.objects.donations(end_date=self.start_date).aggregate(Sum('amount'))['amount__sum']
+        all_reinvestments = AdminReinvestment.objects.reinvestments(end_date=self.start_date).aggregate(Sum('amount'))['amount__sum']
+        all_repayments = AdminRepayment.objects.repayments(end_date=self.start_date).aggregate(Sum('amount'))['amount__sum']
+        cash_in_adjustments = AdminAdjustment.objects.adjustments(end_date=self.start_date, cash_type='cash_in').aggregate(Sum('amount'))['amount__sum']
+        cash_out_adjustments = AdminAdjustment.objects.adjustments(end_date=self.start_date, cash_type='cash_out').aggregate(Sum('amount'))['amount__sum']
+
+        if not all_donations:
```
noahsark769: These aggregations will be None if there is nothing to aggregate over, I think. So you can actually just do `all_donations = all_donations or 0`

revolv/payments/utils.py
```diff
@@ -1,6 +1,398 @@
...
+            cash_in_adjustments = 0
+        if not cash_out_adjustments:
+            cash_out_adjustments = 0
+
+        cash_in = (all_donations * (1 - self.payment_service_fee_rate)) + (all_repayments * (1 - self.revolv_earnings_rate)) + cash_in_adjustments
+        cash_out  = cash_out_adjustments + all_reinvestments
+        net = cash_in - cash_out
+        return net
+
+    def clean_project_name(self, name):
```
noahsark769: Again, don't think you should need this function at all

revolv/payments/utils.py
```diff
@@ -1,6 +1,398 @@
...
+        :return:
+            A boolean of whether the project has associated payment objects.
+        """
+        donations = Payment.objects.donations(project=project, start_date=self.start_date, end_date=self.end_date).count()
+        reinvestments = AdminReinvestment.objects.reinvestments(project=project, start_date=self.start_date, end_date=self.end_date).count()
+        repayments = AdminRepayment.objects.repayments(project=project, start_date=self.start_date, end_date=self.end_date).count()
+
+        return (donations + reinvestments + repayments) != 0
+
+    def move_one_month_forward(self, date):
```
noahsark769: This doesn't use `self` at all, so it should really go in a separate function, not a method of this class

revolv/payments/utils.py
```diff
@@ -1,6 +1,398 @@
...
+        filtered_dict['date_info'] = {}
+        filtered_dict['cash_balances'] = {'beginning_cash_balance': [], 'change_in_cash': [], 'final_cash_balance': [], 'net_cash_in': [], 'net_cash_out': []}
+        date_info = filtered_dict['date_info']
+        date_info['start_date'] = self.start_date
+        date_info['end_date'] = self.end_date
+        date_info['date_list'] = []
+        filtered_dict['project_names'] = []
+        project_names = filtered_dict['project_names']
+        for proj in self.project_filter:
+            if (self.has_payments(proj)):
```
noahsark769: Don't need outer parens

revolv/payments/utils.py
```diff
@@ -1,6 +1,398 @@
...
+                project_names.append(proj.title)
+
+        for ct in ['cash_in', 'cash_out']:
```
noahsark769: how about `cash_type` instead of `ct`? Verbosity is okay if it makes reading the code easier :)

revolv/payments/utils.py
```diff
@@ -1,6 +1,398 @@
...
+                    datetime.datetime object,
+                    datetime.datetime object,
+                    datetime.datetime object,
+                ],
+                'start_date': datetime.datetime object,
+                'end_date': datetime.datetime object,
+            },
+            'project_names': ['Power Community Dance Studio'],
+            'cash_balances': {
+                "net_cash_out": [50, 0, 0],
```
noahsark769: What is the significance of the three elements of this array? I feel like it may be better structured as a dictionary

revolv/static/js/base/accounting.js
```diff
@@ -0,0 +1,78 @@
...
+ * @param {Object} changed_data - An object with either 0 or 1 keys that maps a selector to its new value.
+ */
+function makeAJAXCall(changedData) {
+    var csrftoken = getCookie('csrftoken');
+    $.ajax({
+        url: window.JSONURL,
+        method: "GET",
+        data: changedData,
+        headers: {"X-CSRFToken": csrftoken},
+        success:function(data, textStatus, xhr){
```
noahsark769: One space after the colon

revolv/templates/base/partials/cash_flow_table.html
```diff
@@ -0,0 +1,156 @@
...
+            </tr>
+        </thead>
+        <tbody class="cash-in-body table-body">
+            <tr class="event-type donation">
+                <td class="level-0">Donations</td>
+            </tr>
+            {% for project_name, project_info in data.cash_in.donations.items %}
+            	<tr><td class="level-1">{{ project_name }}</td></tr>
+            	<tr>
+            		<td class="level-2">Total</td>
```
noahsark769: Tabs -> spaces please :)

revolv/templates/base/partials/cash_flow_table.html
```diff
@@ -0,0 +1,156 @@
+{% comment %}
```
noahsark769: Oh man, this file is so much better than the jQuery we had before. I like!

revolv/templates/donor/accounting.html
```diff
@@ -0,0 +1,49 @@
+{% extends "base/base.html" %}
+{% load staticfiles %}
+{% load widget_tweaks %}
+
+{% block head %}
+    <link rel="stylesheet" type="text/css" href="{% static "accounting.css" %}" />
+    <link href='https://fonts.googleapis.com/css?family=Lato:300,500,800' rel='stylesheet' type='text/css'/>
+    <script type="text/javascript">
+        window.JSONURL = "{{ jsonurl }}";
```
noahsark769: It's no longer a JSON url, so we should name it something different

noahsark769:
Made a lot of comments - I think the biggest thing still missing is that you should be querying the projects by id rather than title (so no need to do all that weird project name cleaning and stuff).

Also, the Accounting aggregator is in dire need of tests D:

revolv/payments/utils.py
sameerav: this is just an example that happens to have 3 items oops! there's no significance to the number 3

i'll make that more clear

sameerav:
@noahsark769 is this issue blocked until we fix the payment model to account for payment fees and revolv earnings/is that finished?

noahsark769:
@sameerav not technically, I guess we could just leave a space for it to be filled in, but we should do other priority issues first: [https://github.com/calblueprint/revolv/labels/priority-1](https://github.com/calblueprint/revolv/labels/priority-1)

In particular, #255 shouldn't be super hard and I won't have time to do it until Thursday, so you could get started on that maybe!

noahsark769:
Wow, I forgot how pedantic I used to be on code reviews :/

Anyways, four years later, this repo is no longer the canonical RE-volv repo. Going to close this to get it out of my Pull Request list :)

PR #312, opened 2015-05-29
Fixed signup subscribed to newsletter error

### LINKED ISSUE #303: Can't sign up if newsletter box is unchecked
On staging, you can't create an account if the newsletter box is unchecked: it just redirects to the signup page again with no error messages.

@erictshen do you know why this might be?

erictshen:
Fixes #303 

Fixes bugs where users can't signup if they don't subscribe to the newsletter.

Simply changed required to be False for that field.

http://stackoverflow.com/questions/16955256/booleanfield-default-value-not-set-when-creating-a-model-instance


noahsark769:
Woah, nice! Will merge this when codeship passes.

noahsark769:
SHIP IT

Issue #308, opened 2015-05-29
Share buttons in the donation confirmation modal


noahsark769:
"I just donated, won't you?" is a good way to get the word out about projects and engage donors more.

PR #333, opened 2015-06-03
Sameera donation modal share buttons

sameerav:
super short pull request to add share buttons to the donation modal - addresses #308 
review @noahsark769 ?

### DIFF
--- a/revolv/static/main.css
+++ b/revolv/static/main.css
@@ -5642,6 +5642,10 @@ th.hide-for-touch {
       padding-bottom: 0.9375rem; }
     .modal-content .modal-section.table {
       padding: 0; }
+    .modal-content .modal-section.share-buttons {
+      margin-bottom: 10px; }
+      .modal-content .modal-section.share-buttons .share-button-table-cell {
+        text-align: center; }
   .modal-content .modal-footer-button-group {
     list-style: none;
     margin: 0; }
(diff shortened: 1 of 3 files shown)


noahsark769:
@sameerav can I get a screen cap?

sameerav:
screen shot 2015-06-03 at 5 31 38 pm

my bad, here it is

noahsark769:
So what happens when you click on the Facebook button? What exactly are you sharing a link to, the project page? Also what does the headline if the link say when you share it? The guys from IDEO were saying how important that was so I figure we should make sure it works

sameerav:
screen shot 2015-06-03 at 5 52 54 pm

it does something like this, opens up a window with the link to the project page.
What do you mean by headline?

noahsark769:
By headline, I mean the place where "127.0.0.1" is right now. The sharing isn't get effective if all the Facebook post days is the IP address haha. Is there a way to change that?

noahsark769:
Says*

sameerav:
Hm I think when its on the actual site it'll be a link to the revolv site, but yeah I agree that should be changed to the "I donated, why don't you?" that's in the issue title or something like that

I'll look into it!

noahsark769:
Yup, I bet "I donated to <project name>, won't you?" would be a good one

noahsark769:
Four years later, this repo is no longer the canonical RE-volv repo. Going to close this to get it out of my Pull Request list. Also, hi @sameerav!

Issue #310, opened 2015-05-29
Frontend for associating users with preferred categories


noahsark769:
This is where the magic happens, the donors need to be able to vote on their preferences in the dashboard!

noahsark769:
Fixed by #334

PR #334, opened 2015-06-05
Sameera frontend category preferences

sameerav:
small screenshot:
screen shot 2015-06-05 at 3 42 18 am
medium screenshot:
screen shot 2015-06-05 at 3 42 06 am
large screenshot with buttons selected: 
screen shot 2015-06-05 at 3 42 38 am

I have no idea why this is happening but when we are  viewing from a donor view the categories don't show up for some reason (this happens):
screen shot 2015-06-05 at 3 45 10 am
Do you know why @noahsark769 ?


noahsark769:
Hm, will look into the issue shortly. Also gunna review the code! One style fix: on medium screens, can we have the "What's going on" block be the same width as the category buttons? Also, any reason for no checkmark on medium screens?

revolv/base/forms.py
```diff
@@ -21,6 +21,7 @@ def save(self, commit=True):
         last names.
         """
         user = super(SignupForm, self).save(commit=False)
+        user.save()
```
noahsark769: Why add this? User.save() is not supposed to happen unless `commit` is true, which is down on line 29

revolv/static/js/base/dashboard.js
```diff
@@ -20,6 +20,73 @@ $(document).ready(function () {
...
+        for (var i = 0; i < allCategories.length; i += 1) {
+            var categoryText = $(allCategories[i]).children().last().text();
+            dict[categoryText] = false;
+        }
+
+        for (var i = 0; i < allActiveCategories.length; i += 1) {
+            var categoryText = $(allActiveCategories[i]).children().last().text();
+            dict[categoryText] = true;
+        }
+        makeAJAXCall(dict);
```
noahsark769: I think this should be called something more specific, because we may eventually make many ajax calls via the dashboard. Alternatively, this looks like the only place you call `makeAJAXCall`, so it might be okay to just put the `$.ajax` here: it's not super long or anything.

noahsark769: Also, can we toggle back the active class and show an alert or something if the ajax call fails?

revolv/static/js/base/dashboard.js
```diff
@@ -20,6 +20,73 @@ $(document).ready(function () {
             $this.addClass("fa-rotate-180");
         }
     });
+    
+    /**
+     * This function defines what happens when the different categories are clicked in the
+     * 'my impact' section on the dashboard. The category that is clicked is made active, 
+     * and then an AJAX request is made. The view this AJAX reqest is sent to updates the
+     * preferred_categories attribute of this user.
+     *
```
noahsark769: This is a fantastic docstring! We don't need this blank line though

revolv/static/js/base/dashboard.js
```diff
@@ -20,6 +20,73 @@ $(document).ready(function () {
...
+     * preferred_categories attribute of this user.
+     * 
+     */
+    $(".category-option-container").click(function() {
+        $(this).toggleClass("active");
+        var allActiveCategories = $(".category-option-container.active");
+        var allCategories = $(".category-option-container");
+        var dict = {};
+        for (var i = 0; i < allCategories.length; i += 1) {
+            var categoryText = $(allCategories[i]).children().last().text();
```
noahsark769: Can you use [.find()](https://api.jquery.com/find/) instead of grabbing the last child? If we changed the markup, then this would fail in its current state. Also, can we use the category id (perhaps as a [data](https://api.jquery.com/jquery.data/) attribute) instead of the text? In general, always better to specify ids rather than titles, because ids will never change.

revolv/static/js/base/dashboard.js
```diff
@@ -20,6 +20,73 @@ $(document).ready(function () {
...
+        $.ajax({
+            url: window.categoryURL,
+            method: "GET",
+            data: changedData,
+            headers: {"X-CSRFToken": csrftoken},
+        });
+    }
+
+    /**
+     * Copied from the Django documentation. Checks if a method needs a csrf token.
```
noahsark769: Can you include the url of the django documentation page where this came from? I've used this before too haha, but it's always nice to have a url for someone who hasn't to better be able to understand

revolv/base/views.py
```diff
@@ -59,9 +60,25 @@ def get_context_data(self, **kwargs):
         context["project_dict"] = project_dict
         context["role"] = self.role or "donor"
 
+        context['category_setter_url'] = reverse('dashboard_category_setter')
+        context['categories'] = Category.objects.all().order_by('title')
+        context['preferred_categories'] = RevolvUserProfile.objects.get(user=self.user).preferred_categories.all()
```
noahsark769: In a UserDataMixin, you can just use `self.profile` to get the user's RevolvUserProfile

revolv/base/views.py
```diff
@@ -59,9 +60,25 @@ def get_context_data(self, **kwargs):
...
+
         # TODO (noah): add in support for autoshowing a project based on the active_project GET parameter
         return context
 
+class CategoryPreferenceSetterView(UserDataMixin, View):
```
noahsark769: I think this class should only allow POST right?

revolv/base/views.py
```diff
@@ -59,9 +60,25 @@ def get_context_data(self, **kwargs):
...
+    
+    def dispatch(self, request, *args, **kwargs):
+        standard_response = super(CategoryPreferenceSetterView, self).dispatch(request, *args, **kwargs)
+        user = self.user_profile
+        user.preferred_categories.clear()
+        info_dict = request.GET.dict()
+        for category_string in info_dict:
+            if info_dict[category_string] == 'true':
```
noahsark769: Yup, we should do this based on id, not title, because titles are not unique. Also, comparing to "true" seems a bit hacky: shouldn't we just pass in an array as a parameter? (To do this, see `getlist` in the docs for [QueryDict](https://docs.djangoproject.com/en/1.7/ref/request-response/))

revolv/base/forms.py
sameerav: something about the many to many relationships of the preferred_categories required that the user is saved before setting its attributes (I don't remember exactly)
would this work?
screen shot 2015-06-05 at 2 28 40 pm

revolv/static/dashboard.scss
```diff
@@ -401,6 +401,103 @@ $sidebar-icon-gray: #999;
...
+                margin: 30px 15px;
+                font-size: 1.1rem;
+            }
+
+            .category-option-container {
+                width: 75%;
+                height: 40px;
+                border: 1px solid $light-gray;
+                margin: 5px auto 0 auto;
+                transition: all 0.3s;
```
noahsark769: Can we make a mixin for the transition? Like in mixins.scss

noahsark769: Also should we be doing all here? Or just color? Not sure if width transitions are necessary haha

revolv/base/views.py
```diff
@@ -59,9 +60,25 @@ def get_context_data(self, **kwargs):
         context["project_dict"] = project_dict
         context["role"] = self.role or "donor"
 
+        context['category_setter_url'] = reverse('dashboard_category_setter')
+        context['categories'] = Category.objects.all().order_by('title')
```
noahsark769: This is why the categories don't show up: this is the base view only for the administrator and ambassador dashboards (as the docstring says). You'll need to add the context data to the donor dashboard view too (I think it's in donor/views)

revolv/base/forms.py
noahsark769: Hm, what's the error? If the user needs to be saved before setting its attributes, then how did it ever work before?

noahsark769:
For reference: this is aiming to fix #310

noahsark769:
Also, I think we need a quick test for the category preference setting view!

noahsark769:
(Don't worry about getting this done before the meeting with Andreas, it's in 45 minutes and there's still a fair bit to do haha)

revolv/static/js/base/dashboard.js
sameerav: The category objects are passed in as context data in the view anyway (and that's where the names are generated from) so selecting by name shouldn't cause a problem even if they change

revolv/base/forms.py
sameerav: oops my bad there was actually an error somewhere else so i just reverted this back to how it was before

noahsark769: Yay!

revolv/base/views.py
sameerav: doesn't ajax only take dictionaries as parameters? I made it a little cleaner though!

revolv/static/dashboard.scss
sameerav: LOL true but the width isn't gonna change right so it won't matter?

revolv/static/js/base/dashboard.js
noahsark769: Hm, I guess you're right. Do you think it would be better to use IDs? I think it would be more robust that way, but if you don't think so I don't think it's a big deal

revolv/base/views.py
noahsark769: Nah you can pass arrays by setting different values to the same key a lot of times. Or you could comma separate them

revolv/static/dashboard.scss
noahsark769: It might if the page resizes right? Or if we had a really long category name which went on two lines at small sizes, the width would animate? Also isn't animating everything going to make it slower? I thought I heard that somewhere

revolv/static/js/base/dashboard.js
sameerav: i changed it anyway because you're right its more robust and it made code simpler anyway

revolv/base/tests/test_views.py
```diff
@@ -1,9 +1,33 @@
...
+class CategorySetterTestCase(TestUserMixin, UserTestingMixin, TestCase):
+    
+    def test_category_setting(self):
+        self.test_profile.make_administrator()
+        response = self.send_test_user_login_request()
+
+        # checks to see that the user has no pre-existing preferences
+        self.assertEqual(len(self.test_profile.preferred_categories.all()), 0)
+        
+        self.client.post(reverse('dashboard_category_setter'), {'1': '', '2': ''}, HTTP_X_REQUESTED_WITH='XMLHttpRequest')
```
noahsark769: Instead of hardcoding the ids like this, can you just grab the first two categories (i.e. `Categories.objects.all()[:2]`) and use their `.id` attribute instead? This way, if we every have to delete and reload categories into the database, this test will still pass, even though the ids will be different

revolv/base/views.py
noahsark769: I think you still need `http_methods` or whatever here

revolv/static/js/base/dashboard.js
```diff
@@ -20,6 +20,66 @@ $(document).ready(function () {
...
+
+        var csrftoken = getCookie('csrftoken');
+        $.ajax({
+            url: window.categoryURL,
+            method: "POST",
+            data: categoryData,
+            headers: {"X-CSRFToken": csrftoken},
+            error: function (xhr, textStatus, thrownError){
+                $(this).toggleClass("active");
+                alert("Sorry, something went wrong! Try again?");
```
noahsark769: "Please try again." would be nicer haha

revolv/templates/base/partials/category_preferences.html
```diff
@@ -0,0 +1,17 @@
+{% load staticfiles %}
+
+<div class="large-1 columns gray-bar-container">
+    <div class="gray-bar"></div>
+</div>
+<div class="large-11 columns choice-container">
+    <div class="choice-header">What type of projects should we invest your repayments in next?</div>
+    {% for category in categories %}
+        <div class="category-option-container {% if category in preferred_categories %}active{% endif %}" id="{{ category.id }}">
```
noahsark769: Not a super good idea to put the category.id in the id of the element, since if we did this for something else on the page also, then there would be two elements on the same page with the same id, which is not supposed to happen and can sometimes create problems for javascript (esp when trying to find elements by id). Better idea would be to do `data-category-id={{category.id}}`: you can this access it with jQuery by doing `$(...).data("category-id")`; Read more at [http://stackoverflow.com/questions/7261619/jquery-data-vs-attr](http://stackoverflow.com/questions/7261619/jquery-data-vs-attr)

noahsark769:
Okay as for why codeship is failing...I have no idea. It's failing in master too, on a branch that passed yesterday, so it looks like something changed on PayPal's side. It looks like [PayPal Rest SDK is now at version 1.9.0](https://crate.io/packages/paypalrestsdk/1.9.0/) and we have 1.4.1 installed, so can you maybe try upping the version in requirements.txt to 1.9.0 and seeing if that fixes it?

noahsark769:
@aykamko maybe you can chime in about the paypal stuff...

noahsark769:
Huh it looks like the paypal sdk might just be down lol

sameerav:
@noahsark769 everything should be fixed now except the paypal stuff which is still going on

revolv/base/tests/test_views.py
```diff
@@ -1,9 +1,40 @@
...
+from django.core.urlresolvers import reverse
 from django.test import TestCase
 from revolv.base.models import RevolvUserProfile
 from revolv.lib.testing import TestUserMixin, UserTestingMixin
+from revolv.project.models import Category
 
 
+class CategorySetterTestCase(TestUserMixin, UserTestingMixin, TestCase):
+    
+    def test_category_setting(self):
```
noahsark769: One last thing: can you put a docstring about what this test is testing here?

noahsark769:
One docstring and one http_methods and you're good to go I think!

noahsark769:
Sweet, no idea what's going on with the Paypal test right now but it's the only one failing so I'm gunna merge this!

PR #338, opened 2015-06-08
Take out django-cms, put in wagtail

### LINKED ISSUE #298: Use Wagtail instead of django-cms
I was thinking that it might be better to use [Wagtail CMS](https://wagtail.io/) instead of Django-CMS. @vivekraghuram I would like your opinion on this - it seems that Wagtail has a more intuitive interface for editing pages, and especially for customizing the nav bar ([see here](http://www.tivix.com/blog/working-with-wagtail-menus/)).

### LINKED ISSUE #337: A bunch of unused templates
I don't think we need anymore:

1. review_project.html
2. donate.html
3. thanks.html

noahsark769:
Fixes #298, also fixes #337 because it was super easy

Took out django cms and put in wagtail! It works in a pretty cool way actually. You can access the wagtail dashboard at `/cms/`, and every admin can log into the dashboard (they have to be a superuser in the database, so I changed `make_administrator` to make them superusers - if you want admins to have wagtail editing privs without restarting the vm, you can do `manage.py seed --clear; manage.py seed`).

The CMS has a "root page" to which you can add child pages, and every child page can have child pages as well, and so on. The nav bar is generated by the children of the root page and the immediate children of those pages.

I added some elementary styles for the custom pages, but a few issues which need to be opened after this is merged:

1. I think that image processing/resizing doesn't work correctly right now, though it may have something to do with running on local rather than an actual server.
2. The document upload feature seems to be using local storage rather than s3 storage for some reason. Not sure if this will work on staging or not, but I think we'll need to look into it.

On the bright side, we no longer have to listen to all those RemovedInDjango18Errors when starting up anything!

@vivekraghuram can you review this please?

screen shot 2015-06-07 at 7 37 34 pm
screen shot 2015-06-07 at 7 37 43 pm
screen shot 2015-06-07 at 8 19 17 pm
screen shot 2015-06-07 at 8 19 26 pm


vivekraghuram:
Reviewing... Btw checkout this addon: https://addons.mozilla.org/en-US/firefox/addon/nimbus-screenshot/
Super helpful for screenshotting entire web pages.

requirements.txt
```diff
@@ -22,20 +22,21 @@ django-webtest==1.7.7
...
+# html5lib==0.999
+# six==1.3.0
+# django-filer==0.9.5
+# cmsplugin-filer==0.9.5
+# django-reversion==1.7
+# djangocms-text-ckeditor==2.4.1
+# djangocms-picture==0.1
+# djangocms-googlemap==0.2
+# djangocms-file==0.1
+# djangocms-video==0.1
```
vivekraghuram: Might as well delete. You can always look on github to see previous version stuff.

revolv/project/views.py
```diff
@@ -21,11 +21,11 @@ class DonationLevelFormSetMixin(object):
     the Create Project and Update Project page.
     """
 
-    def get_donation_level_formset(self, extra = 2):
+    def get_donation_level_formset(self, extra=2):
```
vivekraghuram: No spaces around equals sign?

revolv/project/views.py-e
```diff
@@ -0,0 +1,260 @@
+from decimal import Decimal
```
vivekraghuram: I don't believe you want to commit this file...

revolv/templates/404.html
```diff
@@ -1,7 +1,7 @@
 {% extends "base/base.html" %}
 {% load staticfiles %}
 
-{% block title %}Page Not Found{% endblock %}
+{% block title %}Page Not Found (404) |{% endblock %}
```
vivekraghuram: What's the pipe for?

requirements.txt
noahsark769: Oh yeah oops, totally forgot that

vivekraghuram:
Don't really know much about the wagtail parts so can't comment there. Otherwise lgtm pending tests passing.

revolv/project/views.py
noahsark769: Weird, I didn't actually change that...must have been cought by one of the python flakers. I guess we should leave it in, it is PEP8 after all

revolv/project/views.py-e
noahsark769: God dammit these dumb -e files, i don't even know where they come from

revolv/templates/404.html
noahsark769: It's cause it extends base.html, and I couldn't figure out a better way to get the `Page Not Found (404) | RE-volv` in there. What do you think about that format?

noahsark769:
Yeah, this is my first foray into wagtail so not super sure if I'm doing it the right way, but I followed a lot of code from their demo app so at least there's that (all the template tag and menu stuff). Will fix the few comments soon.

As for the tests, the Paypal Test mysteriously started failing with an unknown error, on every branch, the other day. Everything is passing except that, so tests should be good - the paypal error is unrelated to these changes and we're trying to work out what's going on, something on paypal's end probably.

noahsark769:
@vivekraghuram look good now? If so, I'll merge it in the morning or something

revolv/templates/404.html
vivekraghuram: Odd but w/e lol

vivekraghuram:
LGTM

noahsark769:
SWEET I'm just gunna merge it now then! Oh wait actually gunna remove some methods that I just realized were unused

noahsark769:
Alright merge incoming after codeship

noahsark769:
HOKAY only the paypal test failed to merge time!

PR #353, opened 2015-06-18
Serve static assets from s3

### LINKED ISSUE #135: Move static assets to s3
This would make things load faster :)

### LINKED ISSUE #342: Font awesome is broken on the dashboard
This has to do with the fact that we aren't committing bower files. I think I broke this when I was trying to fix #290, but it should actually be fixed if we deployed staticfiles to s3, as in #135.

noahsark769:
Fixes #135 and also fixes #342. Not much to change here actually! @vivekraghuram I updated settings to serve stuff from s3, and I also logged onto the AWS console and updated the bucket settings to be publicly viewable and CORSable from our domains. Note: when pushing from now on, we'll need to add `manage.py collectstatic` in addition to `migrate` in order to copy new static files over to s3 - I'll add this to the codeship build. Can you lgtm this quickly? :D

### DIFF
--- a/revolv/custom_storages.py
+++ b/revolv/custom_storages.py
@@ -26,10 +26,10 @@ from storages.backends.s3boto import S3BotoStorage
 
 
 class RevolvProductionStaticStorage(S3BotoStorage):
-    """Storage that tells boto to collect files into the staging directory in our S3 bucket."""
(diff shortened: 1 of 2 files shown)


vivekraghuram:
lgtm

PR #357, opened 2015-06-18
Two random tiny fixes

noahsark769:
1. Make the `manage.py seed --quiet` arg actually silence the logging it does
2. Remove a damn -e file. I still have no idea where those come from

@vivekraghuram can you lgtm this real quick? sorry for asking for so many reviews :/


vivekraghuram:
lgtm and no worries

PR #358, opened 2015-06-19
autorepay implemented yoooo

### LINKED ISSUE #193: Automatic repayments
We need to have a task to automatically create repayments based on the project's internal rate of return. This task should run every 1st and 15th of the month, and should automatically create a repayment and send all administrators a notification email.

We also need EXTENSIVE DOCUMENTATION on this as it is something that may need to be changed by RE-volv's dev team as it scales.

### LINKED ISSUE #291: Make sure we handle how RE-volv takes money from repayment correctly

aykamko:
Fixes #291 and fixes #193

### DIFF
--- a/revolv/base/management/commands/autorepay.py
+++ b/revolv/base/management/commands/autorepay.py
@@ -1,7 +1,6 @@
-import sys
 from optparse import make_option
 
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
 from django.utils import timezone
 from revolv.payments.models import AdminRepayment
 from revolv.project.models import Project
@@ -32,10 +31,7 @@ class Command(BaseCommand):
         """
         day = timezone.now().day
         if not any([day == 1, day == 15, options['override']]):
-            print >> sys.stderr, ("autorepay command can only be run on the 1st "
-                                  "or 15th of every month unless the `--override` "
-                                  "flag is supplied. use with caution")
-            sys.exit(1)
+            raise CommandError("autorepay command can only be run on the 1st or 15th of every month unless the `--override` flag is supplied. use with caution")
 
         cur_year = timezone.now().year
         for proj in Project.objects.get_completed():
--- a/revolv/base/management/commands/autorepay.py
+++ b/revolv/base/management/commands/autorepay.py
@@ -1,0 +1,49 @@
+import sys
+from optparse import make_option
+
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+from revolv.payments.models import AdminRepayment
+from revolv.project.models import Project
+
+
+class Command(BaseCommand):
+    option_list = BaseCommand.option_list + (
+        make_option('--override',
+                    action='store_true',
+                    dest='override',
+                    default=False,
+                    help='Forcibly run command on day other than 1st or 15th.'),
+    )
+
+    def handle(self, *args, **options):
+        """
+        Run this function from the command line with `python manage.py autorepay`.
+
+        This command iterates through all completed projects and makes a
+        repayments to each of them according to their respective
+        MonthlyReinvestableAmount and MonthlyOrganizationalCost.
+
+        A Repayment is only made towards a project if it hasn't already been
+        fully repayed.
+
+        This command is cron-ed to run on the 1st and 15th of every month. To
+        run it on any other day, the `--override` flag must be supplied.
+        """
+        day = timezone.now().day
+        if not any([day == 1, day == 15, options['override']]):
+            print >> sys.stderr, ("autorepay command can only be run on the 1st "
+                                  "or 15th of every month unless the `--override` "
+                                  "flag is supplied. use with caution")
+            sys.exit(1)
+
+        cur_year = timezone.now().year
+        for proj in Project.objects.get_completed():
+            if not proj.is_repaid:
+                repayment = AdminRepayment(
+                    reinvestable=(proj.monthlyreinvestableamount_set.get(year=cur_year).amount / 2),
+                    organizational_cost=(proj.monthlyorganizationalcost_set.get(year=cur_year).amount / 2),
+                    admin=None,
+                    project=proj
+                )
+                repayment.save()
--- a/revolv/lib/tests/test_autorepay.py
+++ b/revolv/lib/tests/test_autorepay.py
@@ -11,27 +11,33 @@ from revolv.project.models import (MonthlyOrganizationalCost,
 
 class AutoRepayTestCase(TestCase):
     def test_off_day(self):
+        """
+        Tests that the `autorepay` command only runs successfully on the 1st and
+        15th of every month. The command must fail on any other day *unless* the
+        --override flag is given from the command line.
+        """
         with mock.patch.object(timezone, 'now', return_value=datetime.datetime(2016, 3, 4)):
-            err_stream_fail = StringIO()
             try:
-                call_command('autorepay', override=False, stderr=err_stream_fail)
-                self.fail("shouldn't succeed")
-            except SystemExit:
-                pass
+                call_command('autorepay', override=False, stderr=StringIO())
+                self.fail("autorepay command shouldn't succeed on an off day if the `override` flag is not set")
+            except SystemExit as e:
+                self.assertEqual(1, e.code)
 
             call_command('autorepay', override=True)
 
     def test_amount(self):
+        """
+        Tests that the `autorepay` command repays the correct amount. Since the
+        command runs twice per month, on any given run of the command, it should
+        repay all qualified projects by (MonthlyReinvestableAmount / 2).
+        """
         project = Project.factories.completed.create(funding_goal=1200)
-        cur_year = timezone.now().year
-        MonthlyReinvestableAmount(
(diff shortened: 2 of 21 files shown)


revolv/base/management/commands/autorepay.py
```diff
@@ -0,0 +1,49 @@
...
+
+        This command is cron-ed to run on the 1st and 15th of every month. To
+        run it on any other day, the `--override` flag must be supplied.
+        """
+        day = timezone.now().day
+        if not any([day == 1, day == 15, options['override']]):
+            print >> sys.stderr, ("autorepay command can only be run on the 1st "
+                                  "or 15th of every month unless the `--override` "
+                                  "flag is supplied. use with caution")
+            sys.exit(1)
```
noahsark769: You should probably just `return` not sys.exit right?

noahsark769: Or hm well I guess in the test below you caught a `SystemExit`...doesn't it make more sense to return though? Any specific reason for doing sys.exit?

revolv/lib/tests/test_autorepay.py
```diff
@@ -0,0 +1,64 @@
...
+from revolv.project.models import (MonthlyOrganizationalCost,
+                                   MonthlyReinvestableAmount, Project)
+
+
+class AutoRepayTestCase(TestCase):
+    def test_off_day(self):
+        with mock.patch.object(timezone, 'now', return_value=datetime.datetime(2016, 3, 4)):
+            err_stream_fail = StringIO()
+            try:
+                call_command('autorepay', override=False, stderr=err_stream_fail)
```
noahsark769: Woah this is cool I didn't realize you could override stderr in call_command

revolv/lib/tests/test_autorepay.py
```diff
@@ -0,0 +1,64 @@
...
+                self.fail("shouldn't succeed")
```
noahsark769: Should probably have a more descriptive message

revolv/lib/tests/test_autorepay.py
```diff
@@ -0,0 +1,64 @@
...
+import mock
+from django.core.management import call_command
+from django.test import TestCase
+from django.utils import timezone
+from revolv.project.models import (MonthlyOrganizationalCost,
+                                   MonthlyReinvestableAmount, Project)
+
+
+class AutoRepayTestCase(TestCase):
+    def test_off_day(self):
```
noahsark769: These tests could use docstrings

revolv/lib/tests/test_autorepay.py
```diff
@@ -0,0 +1,64 @@
...
+            call_command('autorepay', override=True)
+
+    def test_amount(self):
+        project = Project.factories.completed.create(funding_goal=1200)
+        cur_year = timezone.now().year
+        MonthlyReinvestableAmount(
+            project=project,
+            year=cur_year,
+            amount=200
+        ).save()
```
noahsark769: You can do `.objects.create(...)` instead of `(...).save()`...I don't really know if that's any better though haha, so up to you

revolv/payments/models.py
```diff
@@ -62,6 +63,10 @@ class AdminRepayment(models.Model):
     objects = AdminRepaymentManager()
     factories = ImportProxy("revolv.payments.factories", "AdminRepaymentFactories")
 
+    @property
+    def reinvestable_amount(self):
+        return self.amount - self.organizational_cost
```
noahsark769: Didn't you take out this `amount` member?

revolv/project/migrations/0050_auto_20150615_0435.py
```diff
@@ -0,0 +1,63 @@
...
+            },
+            bases=(models.Model,),
+        ),
+        migrations.AlterField(
+            model_name='donationlevel',
+            name='amount',
+            field=models.DecimalField(max_digits=15, decimal_places=2),
+            preserve_default=True,
+        ),
+        migrations.AlterField(
```
noahsark769: You've merged in the latest master, correct? This update for the RichTextField was also in another migration, and they might conflict?

revolv/project/models.py
```diff
@@ -484,6 +491,12 @@ def donation_levels(self):
         """
         return self.donationlevel_set.all()
 
+    def get_escalator_rate(self, year):
+        """
+        :return: The escalator rate for the given year.
+        """
+        return self.yearlyescalatorrate_set.get(year=year)
```
noahsark769: We don't have yearlyescalatorrate on the model anymore right?

revolv/project/models.py
```diff
@@ -537,4 +550,33 @@ class DonationLevel(models.Model):
...
+    amount = models.DecimalField(
+        max_digits=15,
+        decimal_places=2
+    )
+
+
+class MonthlyReinvestableAmount(models.Model):
+    """
+    Part of the money to be repaid to RE-volv per month for a particular project
+    that can be made available for reinvestment.
```
noahsark769: Here would be a good place to explain how the monthly reinvestable amounts work with respect to the project (e.g. that there exists one monthly reinvestable amount for each year and to calculate what amount should be reinvested, we look at the one for the current year)

noahsark769:
Looks pretty good, just need to update some docs I think. Also there's a [TODO in project.models.py](https://github.com/calblueprint/revolv/blob/master/revolv/project/models.py#L404) (line 404) that should use MonthlyReinvestableAmounts

noahsark769:
@aykamko bump

revolv/base/management/commands/autorepay.py
aykamko: I want `sys.exit(1)` because I'd like for the command to exit with a `FAIL` system code. afaik it's standard behavior for a unix command to exit with a bad exit code if it fails, and this is one of those cases imo

revolv/payments/models.py
aykamko: I did. Which means we probably don't have a test for this; I'll fix that

revolv/lib/tests/test_autorepay.py
aykamko: brainfart, i'll change it

revolv/project/migrations/0050_auto_20150615_0435.py
aykamko: hmm, i thought i did? ill look into it

revolv/project/models.py
aykamko: correct

revolv/base/management/commands/autorepay.py
noahsark769: Hm alright I'm down with that

revolv/payments/models.py
noahsark769: :+1:

revolv/project/migrations/0050_auto_20150615_0435.py
noahsark769: I think the duplicate is in [https://github.com/calblueprint/revolv/blob/master/revolv/project/migrations/0049_auto_20150615_0053.py](https://github.com/calblueprint/revolv/blob/master/revolv/project/migrations/0049_auto_20150615_0053.py)

revolv/project/models.py
noahsark769: Excellent then we don't need this :)

revolv/lib/tests/test_autorepay.py
aykamko: done

noahsark769:
@aykamko what's the status on this? Did you ever implement the change to allow projects not to accept repayments (i.e. the test project)?

PR #360, opened 2015-06-22
TOS blurb on the signin page

### LINKED ISSUE #304: Signup page checkbox for agreeing to TOS
We need this, but we also need to have a TOS page first. Will need wait on RE-volv to draft that up, I think.

noahsark769:
Fixes #304 

Just added a CMS setting for a blurb about the TOS, pretty simple. The link in the default for this field might not actually exist, but that's okay because it's CMS customizable.

@vivekraghuram can you give this a quick look?

screen shot 2015-06-21 at 7 25 00 pm


vivekraghuram:
lgtm although this could be a little weird if they don't link anything but they probably will

PR #368, opened 2015-06-26
Donor stats table on dashboard project page

### LINKED ISSUE #321: [Dashboard Statistic] Donor stats table
Implement the table that has donors, in this mockup, total days (maybe "timeline" is better), and days past in this mockup: new mockup

noahsark769:
Fixes #321 

Added the donor stats table! Also added the notion of a project's start date, which is determined by when it becomes approved.

@sameerav can you review this?

screen shot 2015-06-26 at 5 17 04 pm
screen shot 2015-06-26 at 5 17 14 pm


revolv/base/management/commands/seed.py
```diff
@@ -99,7 +99,8 @@ class ProjectSeedSpec(SeedSpec):
         "location": "2415 Bowditch St, Berkeley, CA 94704, United States",
         "location_latitude": 37.8670289,
         "location_longitude": -122.2561597,
-        "end_date": datetime.date(2050, 10, 8),
+        "start_date": datetime.date(2014, 1, 1),
+        "end_date": datetime.date(2015, 1, 1),
```
sameerav: should we make this datetime datetime.datetime.now

revolv/base/management/commands/seed.py
```diff
@@ -119,7 +120,8 @@ class ProjectSeedSpec(SeedSpec):
         "location": "1238 5th Street, Berkeley, CA, United States",
         "location_latitude": 37.87968940000000,
         "location_longitude": -122.30289330000000,
-        "end_date": datetime.date(2175, 1, 1),
+        "start_date": datetime.date(2014, 1, 1),
+        "end_date": datetime.date(2015, 1, 1),
```
sameerav: same with this end date

revolv/base/management/commands/seed.py
noahsark769: Nah, I think it's good to have projects that are in the middle of running, because then we can see days left, total days, days so far as different numbers so as not to get confused between them.

I actually think the end date should be relative though, so we don't have to update it. I will put that in now

revolv/project/models.py
```diff
@@ -481,14 +490,56 @@ def partial_repayment_as_js(self):
...
+            return None
+        difference = (datetime.date.today() - self.start_date).days
+        if difference < 0:
+            return 0
+        if difference > self.total_days:
+            return self.total_days
+        return difference
 
     @property
     def days_left(self):
```
sameerav: isn't this method the same as days_until_end?

revolv/base/management/commands/seed.py
noahsark769: This one is important to keep as it is, since the project is supposed to be completed, and the date should thus be in the past.

revolv/project/models.py
```diff
@@ -481,14 +490,56 @@ def partial_repayment_as_js(self):
...
+        """
+        :return: the integer number of days until the end of this project,
+        or 0 if the project's campaign has finished.
+        """
         return max(self.days_until_end, 0)
 
     def formatted_days_left(self):
+        """
```
sameerav: if this is being used in a template, wouldn't it be better to format this in the template?

noahsark769: Yeeaaahhh technically. But not really related to this PR, I just added a docstring because I realized it didn't have one. Will add a TODO though, because you're right that should technically be a template tag or template partial or something

revolv/static/dashboard.scss
```diff
@@ -358,6 +358,12 @@ $sidebar-icon-gray: #999;
 
         .donation-stats {
             margin-bottom: 40px;
+
+            // make sure all children are aligned vertically to the center
+            & > * {
```
sameerav: woah i did not know this was possible

revolv/static/dashboard.scss
```diff
@@ -373,6 +379,30 @@ $sidebar-icon-gray: #999;
                     font-weight: 300;
                 }
             }
+
+            .donor-stats-table {
+                .donor-stats-table-row {
+                    border-bottom: 1px solid #eee;
```
sameerav: #eee should probably be a variable because it's used a couple times :)

noahsark769: You are correct

noahsark769:
@sameerav addressing your changes in real time :D

noahsark769:
@sameerav bump

sameerav:
@noahsark769 lgtm!

PR #379, opened 2015-07-21
Add a "STAGED" project status

### LINKED ISSUE #255: Add a project status: STAGED

noahsark769:
Fixes #255 

@erictshen can you review this?


revolv/project/models.py
```diff
@@ -291,17 +325,18 @@ def has_owner(self, ambassador):
 
     def approve_project(self):
         self.project_status = Project.ACTIVE
-        self.start_date = datetime.date.today()
+        if self.start_date is None:
```
erictshen: Good call. Would we need to do something like reset the end_date when the project is completed?

noahsark769: Don't think so, since end_date can technically be edited by the project edit page. Basically we need this because you can EITHER select a start date via the edit page OR leave it blank and have it auto-fill to be the date you press "approve"

PR #384, opened 2015-07-22
Remove geographic category 

### LINKED ISSUE #376: Take out the "geographic" default category
Should be a simple data migration change

vivekraghuram:
Can't remember if I did this the proper django way. Removed the category so the original data migration that creates them shouldn't event make the geographic category. The new data migration should remove it from any existing databases. Resolves #376 

@noahsark769 sí?


noahsark769:
Looks great! @vivekraghuram++

PR #385, opened 2015-07-23
Homepage animations

### LINKED ISSUE #315: Big empty space on homepage when there is no project featured
See @andylouisqin's comment [here](https://github.com/calblueprint/revolv/pull/294#issuecomment-107110616). Should note also that this messes up the "learn about RE-volv" animation a lot, since scrolling offsets are wrong.

sameerav:
I fixed the animation on the homepage when there's no featured project, which got rid of the space at the  bottom. Not sure if this is a good way to do this though.

Could you take a look @noahsark769 ?

Fixes #315


revolv/templates/base/animations/nav_spine.html
```diff
@@ -28,4 +29,37 @@
...
\ No newline at end of file
+</div>
+{% else %}
+<div 
+    data-300="margin-left: -100px;"
+    data-700="margin-left: 100px;"
+    data-1750="margin-left: 100px;"
+    data-1850="margin-left: -100px;"
+    class="nav-spine">
+        <div class="first nav-spine-item">1</div>
```
noahsark769: The actual text here is duplicated from above, even though the offsets are different. I think it would be good to have every offset group in a specific if block by itself: this way, even though there's a lot of if/else logic, we're not duplicating strings, so the duplication only happens where it needs to.

noahsark769:
@sameerav this looks fine to me, the one comment I made should be pretty quick to fix. Nice!

sameerav:
@noahsark769 finished!

noahsark769:
Awesome! Thanks for doing this @sameerav! Merging now!

PR #396, opened 2016-02-04
Content page models & styles

nmashton:
This adds numerous improvements to the layout and style of "REvolv custom pages", which are the generic CMS content pages for the REvolv site.

Generic improvements:
- On content pages, the footer is made sticky.
- The margin at the top of the main content well on content pages is made screen-width-sensitive to accommodate the changes in the height of the header on different widths.
- Some padding is added to improve appearance.

The more substantial improvement comes about from replacing the content page's simple RichTextField-based content with a more sophisticated StreamField that opens up a number of possibilities. The StreamField is a means of editing page content as a series of "blocks". It is documented fully [here](http://docs.wagtail.io/en/v1.3.1/topics/streamfield.html).

This pull request makes the following StreamField-related additions to the models and templates:
- The RevolvCustomPage model's `body` attribute is turned into a StreamField that accepts three block types: rich text, an ImageBlock (see blow), and a RichListBlock (see below).
- A custom ImageBlock block. This block includes an image; a "size" parameter (allowing the user to choose an Extra Small, Small, etc image); and a "layout" parameter (allowing the user to float the image left and right or make it inline or whatever).
- A custom RichListBlock block. This is a ListBlock where each list item is, basically, a StreamBlock that can contain arbitrarily many ImageBlocks or RichTextBlocks. But that StreamBlock also accepts parameters that define the display type and let the user set `align-items` and `justify-content` values. The upshot of this is that the user can create a row of items that are vertically centered by selecting the `Row` display type; the `Start` main axis value; and the `Center` perpendicular axis value.

A good way to test this would be to recreate the [Solar Ambassadors page](http://re-volv.org/content/solar-ambassadors-0) from the production site. To do this, first run the `grunt sass` build step to build the site CSS. Then:
1. Go to `/cms/`. Go to the _Explorer_ area. Choose a parent page. Press the **Add Child Page** button to create a page. Select **Revolv Custom Page** as the page type.
2. Give the page a title.
3. Under _Body_, you should see a greyed in area offering you three choices, like so: 
   image — choose **Rich Text**. Enter in some text (to emulate the appearance of the lead-in text at the top of the production Solar Ambassadors page).
4. Underneath your newly added rich text section, there will be a grey plus sign. Click this; it will display the same grey selection panel as before. Click **Rich List**.
5. You will see the edit panel for a Rich List and its items: 
   image — select Display Type _Row_ and Perpendicular Axis _Center_.
6. Use the grey panel beneath the select inputs to add an Image. Upload an image using the file picker. Select the size _Small_. Select the Layout type _Block_.
7. Use the grey plus sign to open up the grey panel and add a Rich Text field. Enter in some text (to emulate the text next to photos on the Solar Ambassador prod page).
8. Press **Add Another** to add another row. Repeat steps 5-7 to create as many "solar ambassadors" as you want.
9. At the bottom of the page, where it says **Save as Draft**, use the select arrow to choose **Publish** and publish the page. You can then view it live and verify that it has worked. (Styles will not work properly until you publish.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,6 @@ node_modules/
 celerybeat.pid
 # IDEA
 .idea
+
+# RE-volv dist
+revolv/media/
--- a/revolv/revolv_cms/migrations/0032_auto_20160204_1405.py
+++ b/revolv/revolv_cms/migrations/0032_auto_20160204_1405.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import wagtail.wagtailcore.fields
+import wagtail.wagtailcore.blocks
+import wagtail.wagtailimages.blocks
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('revolv_cms', '0031_auto_20160131_2213'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='revolvcustompage',
+            name='body',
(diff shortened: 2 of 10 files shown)


nmashton:
PR accidentally made to non-forked repo.

